PR #27, opened 2020-01-21
Only use a global bootstrap server

pilhuhn:
A global bootstrap should be enough. We don't need them per topic.

### DIFF
--- a/external/src/main/resources/application.properties
+++ b/external/src/main/resources/application.properties
@@ -9,10 +9,13 @@ kafka-streams.consumer.session.timeout.ms=250
 kafka-streams.consumer.heartbeat.interval.ms=200
 #external.org.hawkular.alerts.ui.path=/home/michael/projects/hawkular-alerts-ui
 
+# Kafka bootstrap applies to all topics
+kafka.bootstrap.servers=localhost:9092
+
 # Source <= hosts
 mp.messaging.incoming.kafka-hosts.connector=smallrye-kafka
 mp.messaging.incoming.kafka-hosts.topic=hosts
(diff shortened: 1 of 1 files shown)


external/src/main/resources/application.properties
```diff
@@ -9,23 +9,26 @@ kafka-streams.consumer.session.timeout.ms=250
...
+mp.messaging.incoming.kafka-hosts.topic=platform.inventory.host-egress
+# mp.messaging.incoming.kafka-hosts.bootstrap.servers=localhost:9092
 mp.messaging.incoming.kafka-hosts.group.id=correlation
 mp.messaging.incoming.kafka-hosts.key.deserializer=org.apache.kafka.common.serialization.LongDeserializer
 mp.messaging.incoming.kafka-hosts.value.deserializer=io.vertx.kafka.client.serialization.JsonObjectDeserializer
 # Source <= hosts
 
 mp.messaging.outgoing.kafka-examples.connector=smallrye-kafka
-mp.messaging.outgoing.kafka-examples.topic=hosts
+mp.messaging.outgoing.kafka-examples.topic=platform.custom-policies.actions-email
```
burmanm: Were these decided? I thought the proposal was actions.email ?

pilhuhn:
On 21 Jan 2020, at 21:41, Michael Burman wrote:

> Were these decided? I thought the proposal was actions.email

I took what Chris put in Slack

external/src/main/resources/application.properties
theute: See the message on Slack, they preferred to keep 3 parts separated by "." and then changed the last . with a -

PR #31, opened 2020-02-05
[CPOL-70] Added metrics endpoint

theute:
Very simple addition so that we can start adding "engine" metrics configuration for Prometheus/Grafana

### DIFF
--- a/external/pom.xml
+++ b/external/pom.xml
@@ -44,6 +44,10 @@
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-metrics</artifactId>
+    </dependency>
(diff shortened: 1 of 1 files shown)


burmanm:
Is this really what the SRE team wants? The microprofile metrics doesn't even expose histogram (the most important metric). Or anything useful really unless we add metrics.

theute:
> Is this really what the SRE team wants? The microprofile metrics doesn't even expose histogram (the most important metric). Or anything useful really unless we add metrics.

This won't be enough, but the point is to be able to configure prometheus instance (and test) already so that when we'll add the necessary metrics we won't have to struggle with that, we'll have some JVM metrics at least...

PR #35, opened 2020-02-20
CPOL-86 CPOL-80 Integrate to beta hooks

burmanm:
Send every event, managed as managed plugin.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/custompolicies/engine/actions/plugins/BetaHooksActionPluginRegister.java
+++ b/external/src/main/java/com/redhat/cloud/custompolicies/engine/actions/plugins/BetaHooksActionPluginRegister.java
@@ -1,0 +1,108 @@
+package com.redhat.cloud.custompolicies.engine.actions.plugins;
+
+import io.smallrye.reactive.messaging.annotations.Channel;
+import io.smallrye.reactive.messaging.annotations.Emitter;
+import io.vertx.axle.core.Vertx;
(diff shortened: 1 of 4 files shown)


external/src/main/java/com/redhat/cloud/custompolicies/engine/actions/plugins/BetaHooksActionPluginRegister.java
```diff
@@ -0,0 +1,108 @@
...
+        return defaultProperties;
+    }
+
+    @Override
+    public void process(ActionMessage msg) throws Exception {
+        // Fields and terminology straight from the target project
+        JsonObject message = new JsonObject();
+        message.put("application", APPLICATION_NAME);
+        message.put("event_type", EVENT_TYPE);
+        message.put("level", 1); // Is this necessary field?
```
theute: I don't think it can be "1" because it needs to match a "level" defined during registration.
I asked on Slack it it should be empty or just not there at all:
https://ansible.slack.com/archives/CFKVBJ6M6/p1582213163012600

theute: So looks like the value should be null according to @adamruzicka 
https://ansible.slack.com/archives/CFKVBJ6M6/p1582281814015800

theute:
Merging and crossing my fingers

PR #36, opened 2020-02-21
Allow the receive to log a short message if an event is received.

pilhuhn:
Also emit timing metrics.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
+++ b/external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
@@ -52,7 +52,7 @@ public class Receiver {
         String tenantId = json.getString(TENANT_ID);
         String insightsId = json.getString(INSIGHT_ID_FIELD);
         if (logEvents) {
-            log.debugf("Got msg for tenant [%s] and insightsId [%s] \n ",  tenantId , insightsId);
+            log.infof("Got msg for tenant [%s] and insightsId [%s] \n ",  tenantId , insightsId);
         }
 
         JsonObject systemProfile = json.getJsonObject("system_profile");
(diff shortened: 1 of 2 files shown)


external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
```diff
@@ -31,16 +32,23 @@
...
     @Inject
     AlertsService alertsService;
 
+    @Timed(name = "engine-incoming", absolute = true)
     @Incoming("kafka-hosts")
     public CompletionStage<Void> processAsync(JsonObject json) {
         String tenantId = json.getString(TENANT_ID);
         String insightsId = json.getString(INSIGHT_ID_FIELD);
+        if (logEvents) {
+            System.out.printf("Got msg for tenant [%s] and insightsId [%s] \n ",  tenantId , insightsId);
```
burmanm: This should use log.debugf()

external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
```diff
@@ -31,16 +32,23 @@
     @ConfigProperty(name = "engine.receiver.store-events")
     boolean storeEvents;
 
+    @ConfigProperty(name = "engine.receiver.log-events", defaultValue = "false")
```
burmanm: Lets keep logging properties on logging configuration.

pilhuhn: How do you mean?
The intention here is to switch this on/off from the env in Kube.

external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
pilhuhn: Well. In container world that all goes to stdout, so I am not sure if that is really needed.

burmanm: It is necessary for the behavior of async processing (System.out isn't async).

external/src/main/java/com/redhat/cloud/custompolicies/engine/process/Receiver.java
burmanm: You can do that with env parameters using ``quarkus.log.category."categories".level``

pilhuhn:
I am merging this, so that we can use it for debugging purposes. We can still change the way the logging is enabled later (or completely remove that again).

PR #46, opened 2020-03-16
Make changes to Receiver error handling

burmanm:
Change kafka-hosts to host-egress in the naming, test that receiver handles errors correctly and add metrics for those situations


external/src/test/java/com/redhat/cloud/custompolicies/engine/ReceiverTest.java
```diff
@@ -16,28 +18,31 @@
...
 
+    @BeforeAll
+    void init() {
+        System.setProperty("hawkular.data", "./target/hawkular.data");
+    }
+
     @Inject
-    @Channel("kafka-hosts")
-    Emitter<JsonObject> hostEmitter;
+    @Channel("host-egress")
```
pilhuhn: What about even calling it inventory.host-egress or so? Just to make it a bit more obvious why something incoming is called 'egress'

burmanm: I'm not sure if  dot is allowed there (the properties file could be messed up).

PR #58, opened 2020-03-27
POL-157 Change hooks.outbox serializer 

burmanm:
Changes serialization to push empty string instead of null field


theute:
I'm not entirely sure that we want empty strings for *all* fields. 
I know we want it for the description, but "maybe" there is part of the payload where we don't want the field at all (I don't think we'll want "null").
We can merge if that's fulfill current requirements though.

burmanm:
Which field in hooks would then be removed? Seems pretty counter-intuitive for processing if some fields can disappear and some can't.

theute:
I agree with your comment, I don't have good examples in mind but I think we had one of those examples.
We can merge and see if the case happen

PR #61, opened 2020-04-01
Changes to updating enable / disable status of trigger

burmanm:
Modified the hashCode and equals to avoid unnecessarily updating the conditions as well as added new interfaces that simplify disable or enable. With some tests on top of that.


external/src/main/java/com/redhat/cloud/policies/engine/metrics/QuarkusRegistryConfig.java
```diff
@@ -0,0 +1,24 @@
...
+
+@ApplicationScoped
+public class QuarkusRegistryConfig {
+
+    public static MsgLogger LOGGER = MsgLogging.getMsgLogger(QuarkusRegistryConfig.class);
+    @Inject
+    Vertx vertx;
+
+    void init(@Observes StartupEvent startupEvent) {
+        LOGGER.infof("=============> VERTX-METRICS: %s", vertx.isMetricsEnabled());
```
pilhuhn, line 20: Technically not really part of this? :)

burmanm: Argh

PR #65, opened 2020-04-17
POL-226 Allow to only retrieve the latest alert per trigger.

pilhuhn:
Without this, getting the lastTriggered datum may return thousands of entries to the client, just to throw most of them away. Limiting by startTime does not help, as we don't know if the alert had fired before that cut-off-time.

A better implementation would be to narrow down the ISPN query in order to only return the latest alert in the first place.


burmanm:
Closing this, superseded by #74 for now (trigger query changes might add the ISPN optimizations for alert fetching if such feature is necessary)

PR #73, opened 2020-04-30
Add support for logging to Sentry

pilhuhn:

### DIFF
--- a/external/pom.xml
+++ b/external/pom.xml
@@ -70,6 +70,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-smallrye-health</artifactId>
     </dependency>
+    <dependency>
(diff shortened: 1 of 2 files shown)


burmanm:
Do we want to expose environment variable?

pilhuhn:
@burmanm Sorry, I don't understand the question. The Env variables (also to switch it on) would be set in Kube/OpenShift

burmanm:
sentry environment, ``quarkus.log.sentry.environment``

pilhuhn:
I forgot to add it, but it does not matter for operational purposes, as it can still be set from ENV or even -D . The values in application.properties are "fallback" values

theute:
Related PR: https://github.com/RedHatInsights/e2e-deploy/pull/1598

PR #82, opened 2020-06-16
POL-293 Provide the real total count to the caller.

pilhuhn:
Bildschirmfoto 2020-06-16 um 12 09 34

### DIFF
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
@@ -557,11 +557,8 @@ public class IspnAlertsServiceImpl implements AlertsService {
             }
             return (Alert) ispnEvent.getEvent();
         }).collect(Collectors.toList());
-//        if (alerts.isEmpty()) {
-//            return new Page<>(alerts, pager, ispnEvents.getTotalSize());
-//        } else {
-            return preparePage(alerts, pager, ispnEvents.getTotalSize());
-//        }
+
+        return preparePage(alerts, pager, ispnEvents.getTotalSize());
     }
 
     // TODO Paging is not properly handled - the returned size is going to be invalid and not all fields are sort supported
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
@@ -546,7 +546,7 @@ public class IspnAlertsServiceImpl implements AlertsService {
            }
         }
 
-        List<IspnEvent> ispnEvents = getEventItems(query, pager);
+        Page<IspnEvent> ispnEvents = getEventItems(query, pager);
         List<Alert> alerts = ispnEvents.stream().map(ispnEvent -> {
             if (criteria != null && criteria.isThin()) {
                 Alert alert = new Alert((Alert) ispnEvent.getEvent());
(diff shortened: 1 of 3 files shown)


pilhuhn:
Ok, that's interesting. CI runs tests with `-Pitests`, which makes all tests pass, unlike my local build without that profile. Which one is correct?

burmanm:
``-Pitest`` should only run more tests, not have any effect in any other case. This PR needs tests to verify the new behavior.

burmanm:
For example, extending ``queryAlertsWithPaging`` and ``queryEventsWithPaging`` should do the work. There's also another implementation that needs to be fixed, which is the Trigger's Paging feature (and the same test for it, ``getTriggersWithPaging``).

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -558,15 +558,15 @@ public Alert getAlert(String tenantId, String alertId, boolean thin) throws Exce
             return (Alert) ispnEvent.getEvent();
         }).collect(Collectors.toList());
         if (alerts.isEmpty()) {
-            return new Page<>(alerts, pager, 0);
+            return new Page<>(alerts, pager, ispnEvents.getTotalSize());
```
burmanm, line 561: What's the difference for these two paths?

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -606,7 +610,9 @@ public Alert getAlert(String tenantId, String alertId, boolean thin) throws Exce
             log.debugf("ParsedQuery: %s, maxResults: %d, startOffset: %d", parsedQuery.getQueryString(), parsedQuery.getMaxResults(), parsedQuery.getStartOffset());
         }
 
-        return parsedQuery.list();
+        List<IspnEvent> list = parsedQuery.list();
+        Page page = new Page(list, pager, totalSize);
```
burmanm, line 614: Nitpicking, but could just be return here - no need for the assignment.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -715,12 +721,12 @@ public Event getEvent(String tenantId, String eventId, boolean thin) throws Exce
             }
         }
 
-        List<IspnEvent> ispnEvents = getEventItems(query, pager);
+        Page<IspnEvent> ispnEvents = getEventItems(query, pager);
         List<Event> events = ispnEvents.stream().map(e -> e.getEvent()).collect(Collectors.toList());
         if (events.isEmpty()) {
-            return new Page<>(events, pager, 0);
+            return new Page<>(events, pager, ispnEvents.getTotalSize());
```
burmanm, line 727: Same question as for the alerts - is there a real difference in these paths? Makes reading code more difficult if one jumps to a function and one doesn't.

burmanm: Also, should we have a test where the return is empty? Two paths or not - it seems to indicate there's a difference in handling them, but that other case isn't tested. Will it break something?

pilhuhn: This is tested when for the 1000 items case it looks at page 20 (of size 50), which is beyond the end.

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -557,16 +557,16 @@ public Alert getAlert(String tenantId, String alertId, boolean thin) throws Exce
             }
             return (Alert) ispnEvent.getEvent();
         }).collect(Collectors.toList());
-        if (alerts.isEmpty()) {
-            return new Page<>(alerts, pager, 0);
-        } else {
-            return preparePage(alerts, pager);
-        }
+//        if (alerts.isEmpty()) {
```
burmanm, line 560: Remove commented code.

PR #89, opened 2020-07-22
Make the sort direction case independent

pilhuhn:


api/src/main/java/org/hawkular/alerts/api/model/paging/Order.java
```diff
@@ -69,13 +69,16 @@ public String toString() {
         }
 
         public static Direction fromShortString(String shortString) {
-            switch (shortString) {
+            if (shortString==null) {
```
burmanm, line 72: Indentation is broken around `==`.

PR #92, opened 2020-07-27
POL-307 Switch to 'platform.inventory.events' topic

pilhuhn:
If the payload says it is a delete event, then skip it.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
@@ -29,7 +29,7 @@ import static org.hawkular.alerts.api.util.Util.isEmpty;
 public class Receiver {
     private final MsgLogger log = MsgLogging.getMsgLogger(Receiver.class);
 
-    public static final String INSIGHTS_REPORT_DATA_ID = "platform.inventory.host-egress";
+    public static final String INSIGHTS_REPORT_DATA_ID = "platform.inventory.events";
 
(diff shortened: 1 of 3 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
```diff
@@ -59,19 +60,34 @@
...
-                    if(log.isTraceEnabled()) {
+                    if (log.isTraceEnabled()) {
                         log.tracef("Received message, input payload: %s", input.getPayload());
                     }
                     JsonObject json = new JsonObject(input.getPayload());
                     return json;
                 }).thenApplyAsync(json -> {
+                    if (json.containsKey(TYPE_FIELD)) {
+                        String eventType = json.getString(TYPE_FIELD);
+                        if (eventType.equals("delete")) {
```
theute, line 78: Should we do a !eventType.equals("insert") here ? (I don't know the actual name of the event) In case they add more types ?

pilhuhn: Wouldn't we miss the 1st report for a new system then?

theute: I guess we need to listen to "created" and "updated" and ignore the rest (only deleted now) ?
Maybe I'm just extra cautious, I'm afraid a new type gets added and we choke on it

pilhuhn: Ok.

PR #102, opened 2020-09-10
Expose metrics from /proc/self/status

pilhuhn:
This is the same code as in ui-backend.


burmanm:
Indentation is wrong (use the IntelliJ here as well). Also, shouldn't manifest.txt be updated when adding new dependency?

pilhuhn:
Manifest work is on hold at the moment, and a central solution will be created. But I have updated it anyway.

PR #106, opened 2020-09-18
POC of a dedicated history handler

pilhuhn:
This is a PoC and not yet complete (the preparePage functionality is missing).
This reduces the data serialized and sent to the UI-backend massively, both speeding up the engine response and the processing in the UI-backend


burmanm:
I don't think this is the right approach, it's not generic enough. If you want thinner alert, then modifying the thin-alert to be thinner should be the way - not to push policies-ui stuff to engine.

pilhuhn:
Perhaps we should start exposing such stuff as GraphQL queries.

Wrt thin alert: I would expect that there will be different use cases where the definition of 'thin' changes.

burmanm:
A projection of some sort could be written. Could be a selection of fields for example.

PR #107, opened 2020-09-21
POC of returning trigger history items only

pilhuhn:
This is a different variant of #106 which mimics the 'thin=true' behaviour to (mostly) return history related items. 
Some of the simple fields could still be nulled out to speed up serialisation to JSON a bit more.


burmanm:
Why do you need evalSet at all?

pilhuhn:
Because it has the data about the tags + ctime when the policy has triggered. At least this was the case when the code first started.

pilhuhn:
And then the ctime on the alert itself is probably close enough (they are not the same) as the ctime on the EvalSet (?)

burmanm:
tags should copied to the alert as well? Didn't we change this some time ago..

PR #109, opened 2020-09-22
For thin mode also null-out the trigger.

pilhuhn:


burmanm:
How would you know which trigger caused the alert?

pilhuhn:
By looking at the parameter that I supplied to findAlerts(). :-)

And then still the triggerId should be enough and the other parts could be retrieved by getting the trigger by its api.

burmanm:
I don't think that helps. Say you wish to display all the notifications / alerts for a customer, how could you link from an alert to whatever generated it? Even with two triggerIds, that would be impossible.

Setting trigger to null removes the triggerId.

pilhuhn:
I understand that nulling out the trigger removes the triggerId that is inside. I was more or less (implicitly) suggesting to have the triggerId as a separate field on Alert itself and only fill this when thin=true and not the trigger structure.

And then technically one could use the knowledge about the current impl and determine the triggerId as the 1st part of the Alert.id ;-)

burmanm:
> And then technically one could use the knowledge about the current impl and determine the triggerId as the 1st part of the Alert.id ;-)

True, but this doesn't work if there's an event with multiple stages (such as missing value).

burmanm:
Although this should have been tested also (alerts with thin are tested, but there's no additional test to check out that these parameters are not changed)

PR #114, opened 2020-10-02
POL-365 POL-366 POL-367 POL-368 Fix the tags odd behavior

burmanm:
Since they can return Collection<String> which is not often what we want.


theute:
Tests look good to me

PR #116, opened 2020-10-08
Enable backend statistics and match test properties with main

burmanm:


engine/src/main/resources/ispn-alerting-local.xml
```diff
@@ -26,7 +26,7 @@
     </local-cache>
 
     <!-- Backend for definitions, actions history, events and alerts -->
-    <local-cache name="backend">
+    <local-cache name="backend" statistics="true">
```
pilhuhn, line 29: Would we want to enable statistics for the cache manager as well?
See https://infinispan.org/docs/dev/titles/configuring/configuring.html#statistics

pilhuhn (approved):

engine/src/main/resources/ispn-alerting-local.xml
burmanm: You mean at the cache-container level?

burmanm: It adds some stuff, like: 

```
# HELP vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time Cache container average number of milliseconds for all write operation in this cache container
# TYPE vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time gauge
vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time{node="nina-24401"} 0.0
# HELP vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time_nanos Cache container average number of nanoseconds for all write operation in this cache container
# TYPE vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time_nanos gauge
vendor_cache_manager_hawkular_alerts_cache_container_stats_average_write_time_nanos{node="nina-24401"} 526797.0
```

burmanm:
The test failure is acceptable, it's because of disabling the data-trigger setting.

PR #121, opened 2020-10-22
Update to Quarkus 1.9.0.Final and enable Micrometer

burmanm:
Replaced Microprofile Metrics with Micrometer implementation. This adds multiple new metrics for the Kafka, but also renames certain metrics which will ruin our Grafana boards unless the queried metrics are changed.


pilhuhn:
Do you have a mapping between old and new metric names -- at least for the ones we are currently using?

burmanm:
They've changed like this:

``application_engine_actions_email_processed_total`` -> ``engine_actions_email_processed_aggregated_total{scope="application",}``

So.. Prometheus naming.

PR #127, opened 2020-12-03
POL-341 Use "updated" as the timestamp for the notification

josejulio:
Haven't tested with real data. I'm using `updated` because that's the value that appearas to be used in inventory as "Last Check-in". That value (updated) also appears in inventory API: `https://cloud.redhat.com/api/inventory/v1/hosts/{the-id}`


pilhuhn:
@iphands @jharting Do you know if _update_ messages on the `platform.inventory.events` topic always have the 'updated' field. And if so in what format?

jharting:
@pilhuhn It does and the timestamp is stored as an ISO 8601 string.

Note that due to [RHCLOUD-9545](https://issues.redhat.com/browse/RHCLOUD-9545) it's not 100% reliable currently.

pilhuhn (approved):
Looks good. Thanks!

PR #129, opened 2020-12-18
POL-410 New Policy app name and event type name

theute:


theute:
Not too surprised by the test failures, on my machine it hangs :-/

PR #130, opened 2021-01-18
POL-432 Assign the tenantId early to avoid "TenantId must be not null…

josejulio:
…" error

 - This happened when we added new actions to existing triggers, preventing
   from finishing the process.


pilhuhn (approved):
Good find.

PR #134, opened 2021-01-25
Add explicit metrics for various rejection reasons.

pilhuhn:
There are several reasons to reject an incoming message in the code.
This MR splits them in the metrics, so that we get a better idea on why metrics are rejected.


josejulio (approved):

josejulio:
Changes looks good, there are some testing failures (NPE). I suspect that you have to instantiate the counters manually in the tests.

pilhuhn:
Yeah. 
Interestingly enough there was even a logic bug in the existing test code, that only the new metrics revealed.

PR #136, opened 2021-02-02
More logs for debugging

theute:


pilhuhn:
What are those .factorypath files? I don't think we need them.

theute:
> What are those .factorypath files? I don't think we need them.

What .factorypath files ?  ;-) 

Thanks !

pilhuhn (approved):

PR #139, opened 2021-03-12
POL-433 Secure connections to/from Kafka

josejulio:
This PR provides a sample config that can be used to secure the connection from `policies-engine` and `kafka`.

There is also a script to help to locally validate the setup.


external/pom.xml
```diff
@@ -30,6 +30,12 @@
       <version>${insights-notification-schemas-java.version}</version>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>${version.org.apache.kafka.kafka-clients}</version>
+    </dependency>
```
josejulio, line 37: @pilhuhn Do you know if is fine to bump this version? Quarkus (or some dependency of it) seems to use `2.5.0` and we require `2.7.0` if we want to allow PEM files.

cc: @gwenneg

gwenneg: I just ran into a Quarkus PR about bumping Kafka to 2.7.0, see https://github.com/quarkusio/quarkus/pull/14732. Apparently they had some issues.

gwenneg: Kafka got updated to 2.7.0 in Quarkus. Considering the PR content, I think it's safe to update our version to 2.7.0 while waiting for a Quarkus release, as long as we don't need to run policies-engine in native mode (Quarkus registered three classes for reflection with the update). There's also a regression on Windows (https://issues.apache.org/jira/browse/KAFKA-12190) which should not be a problem.

external/src/test/groovy/org/hawkular/alerts/rest/AbstractQuarkusITestBase.groovy
```diff
@@ -14,7 +14,7 @@ import java.util.concurrent.atomic.AtomicInteger
  */
 abstract class AbstractQuarkusITestBase {
 
-    static baseURI = 'http://127.0.0.1:8084/hawkular/alerts/'
+    static baseURI = 'http://127.0.0.1:8095/hawkular/alerts/'
```
pilhuhn, line 17: We should (not necessarily as part of this PR) let the port be injected by the Quarkus test framework (see e.g. `@QuarkusTest`)

pilhuhn (approved):
I guess that is cool

gwenneg:
@josejulio Since we just bumped Quarkus to `1.13.6.Final`, policies-engine now depends transitively on kafka-clients `2.8.0`.

gwenneg:
That was my first and last time resolving a conflict with the GitHub UI 😄 (I don't really like conflicts resolution based on merge)

josejulio:
> @josejulio Since we just bumped Quarkus to `1.13.6.Final`, policies-engine now depends transitively on kafka-clients [2.7.0](https://github.com/quarkusio/quarkus/blob/1.13.6.Final/bom/application/pom.xml#L145).

Should we remove it from the POM?

gwenneg:
We should probably depend on `quarkus-kafka-client` instead of the Apache dependency.

josejulio:
> We should probably depend on `quarkus-kafka-client` instead of the Apache dependency.

Updated. Also did some tests and it's connecting to kafka using the secure channel.

gwenneg (approved):
LGTM

PR #141, opened 2021-03-26
Introduce maven-wrapper

gwenneg:
Same PR as https://github.com/RedHatInsights/notifications-backend/pull/126 for the same reasons.


josejulio:
The jenkins fail might be related to https://github.com/RedHatInsights/policies-engine/pull/139/commits/dee2e8cb0314ea25d9ab40b9bc0feba4e4abf79e

gwenneg:
Why is that change suddenly required? Was `8084` claimed be another service?

josejulio:
> Why is that change suddenly required? Was `8084` claimed be another service?

I have no idea. I suppose some lib or github actions is using that now?
Anyhow @pilhuhn said that the port should be provided by Quarkus test, so we might need to change that to pick it up from Quarkus and not hardcoding it.

gwenneg:
Changing the port fixed the build indeed, thanks!

gwenneg:
The port would be automatically set by Quarkus if the REST queries were performed with RestAssured.

pilhuhn:
I think we can do something like this by using  `quarkus.http.test-port=0` and then injecting the base URI - see https://quarkus.io/guides/getting-started-testing#injecting-a-uri.

For this PR, it is probably enough to just use a different port and be done.

josejulio (approved):

PR #147, opened 2021-04-06
NOTIF-170 Group policies from the same "checkin"

josejulio:


external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +47,107 @@
...
+    private final ConcurrentSkipListMap<String, PoliciesAction> notifyBuffer = new ConcurrentSkipListMap<>();
 
     @Inject
     @Channel("webhook")
     @OnOverflow(OnOverflow.Strategy.UNBOUNDED_BUFFER)
     Emitter<String> channel;
 
     @Inject
-    @Metric(absolute = true, name = "engine.actions.webhook.processed")
+    @Metric(absolute = true, name = "engine.actions.notifications.processed")
```
josejulio, line 59: @pilhuhn Do you know if is OK to update these or if something breaks?

gwenneg: It will probably break the metric monitoring in Prometheus but metrics can be renamed so it would only be a temporary issue.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +47,107 @@
...
                 // Same behavior as previously with JsonObjectNoNullSerializer with old hooks
                 value = "";
             }
-            payloadBuilder.addTag(tagEntry.getKey(), value);
-        }
 
-        notificationAction.setAccountId(actionMessage.getAction().getTenantId());
+            if (!context.getTags().containsKey(tagEntry.getKey())) {
+                context.getTags().put(tagEntry.getKey(), new HashSet<>());
+            }
```
gwenneg, line 90:
```suggestion
            context.getTags().putIfAbsent(tagEntry.getKey(), new HashSet<>());
```

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +47,107 @@
...
 
-                    payloadBuilder.addTrigger(triggerId, name);
+                    PoliciesAction.Event event = new PoliciesAction.Event();
+                    event.getPayload().setPolicyCondition(eventEval.getCondition().getExpression());
+                    event.getPayload().setPolicyId(trigger.getId());
+                    event.getPayload().setPolicyName(trigger.getName());
+                    event.getPayload().setPolicyDescription(trigger.getDescription());
+
+                    policiesAction.getEvents().add(event);
+                    break;
```
gwenneg, line 113: Did you add this `break` because there can only be one `EventConditionEval` instance in the `evalSet`?

external/src/test/java/com/redhat/cloud/policies/engine/process/ReceiverTest.java
```diff
@@ -207,7 +204,9 @@ void testMoreComplexInput() throws Exception {
         testSubscriber.assertValueCount(1);
 
         Action action = deserializeAction(testSubscriber.values().get(0));
-        List<Map<String, String>> tags = (List<Map<String, String>>) action.getPayload().get("tags");
+        assertEquals(1, action.getEvents().size());
+
+        List<Map<String, String>> tags = (List<Map<String, String>>) action.getContext().get("tags");
```
gwenneg, line 209: I don't see the `multiple values per tags key` thing you introduced being tested here. Could you improve this test a bit to cover that case?

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +47,107 @@
...
-        Action notificationAction = new Action();
-        notificationAction.setEventType(EVENT_TYPE_NAME);
-        notificationAction.setApplication(APP_NAME);
-        notificationAction.setBundle(BUNDLE_NAME);
-        notificationAction.setTimestamp(
-                LocalDateTime.ofInstant(Instant.ofEpochMilli(actionMessage.getAction().getCtime()), ZoneId.systemDefault())
+        PoliciesAction policiesAction = new PoliciesAction();
+        policiesAction.setAccountId(actionMessage.getAction().getTenantId());
+        policiesAction.setTimestamp(
+            LocalDateTime.ofInstant(Instant.ofEpochMilli(actionMessage.getAction().getCtime()), ZoneId.systemDefault())
```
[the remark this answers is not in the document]
josejulio: I'll test with UTC.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/notification/PoliciesAction.java
```diff
@@ -0,0 +1,190 @@
...
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class PoliciesAction {
+    private String accountId;
+    private LocalDateTime timestamp;
+    private Context context;
+    private Set<Event> events;
```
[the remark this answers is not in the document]
josejulio: > `Event` does not implement `equals` and `hashCode`, is that OK?
That's OK.

> If we receive the same event data twice (assuming that's possible...),
Not possible, each entry is a policy that triggered.

The reason I'm using a `Set` rather than a `List` is because I read that `ConcurrentSkipListMap#merge` is NOT guaranteed to be applied once atomically. Does that make sense?

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +47,107 @@
...
+            return existing;
+        });
     }
 
     @Override
     public void flush() {
+        log.fine(() -> String.format("Starting flush of %d email messages", notifyBuffer.size()));
 
+        while (true) {
+            Map.Entry<String, PoliciesAction> notificationEntry = notifyBuffer.pollFirstEntry();
```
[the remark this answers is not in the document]
josejulio: That's OK, we don't really mind the order.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
josejulio: Yes, there is only one.

external/src/test/java/com/redhat/cloud/policies/engine/process/ReceiverTest.java
josejulio: It was already there. I think is tested somewhere else but if not will add a test for it.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
josejulio: I'll use `computeIfAbsent`. Wasn't aware of these methods. Thanks!

external/src/test/java/com/redhat/cloud/policies/engine/process/ReceiverTest.java
josejulio: Is tested in `testMoreComplexInput`.

The content there has the `location=Charmey` and `location=Neuchatel`

gwenneg: Oh sorry, I missed that.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/notification/PoliciesAction.java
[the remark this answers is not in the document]
gwenneg: > The function is NOT guaranteed to be applied once atomically only if the value is not present.

If I understand correctly that sentence from the Javadoc, it means that the mapping function could be executed more than once and therefore it should not have any side-effect (it should be idempotent). I don't see any side-effects in the current function so I think we're safe whether we use a `List` or a `Set`.

Do we need a `Set`? I'm not sure but maybe I missed something. But if we do because it would avoid duplicates, then `equals` and `hashcode` should be implemented in `Event`. Otherwise, the `Set` won't do anything about duplicates.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -44,72 +51,103 @@
     public static final String APP_NAME = "policies";
     public static final String EVENT_TYPE_NAME = "policy-triggered";
 
+    private static final Logger log = Logger.getLogger(NotificationActionPluginListener.class.getName());
+    private final ConcurrentSkipListMap<String, PoliciesAction> notifyBuffer = new ConcurrentSkipListMap<>();
```
gwenneg, line 55: I should have asked that earlier, but why do we need to take care of concurrency here? I would expect a `@Dependent` bean to be short-lived and used by a single thread. Can you tell me more about this class lifecycle and why it is used by multiple threads?

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/notification/PoliciesAction.java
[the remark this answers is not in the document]
josejulio: The default equals implementation the same as a `==` (checking if is the same reference), that works for this use case I think.

The current implementation adds to `existing` what we have in `addition`, if calling it twice with same `addition`, the set will be OK. With a `List` we will have duplicates (elements from addition twice in the list) which is what I'm trying to avoid here.
We don't need to implement equals/hashcode as we know that the data will come only once, so reference checking is fine in this case.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
josejulio: That's a good question, I don't know the answer to it. 
I was repeating a pattern for doing the aggregation found in an older version of the code: https://github.com/RedHatInsights/policies-engine/blob/e4fb24c768e0309332536718e7bea3674fa79a11/external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/EmailActionPluginListener.java#L50

From my understanding, this class is instantiated at code initialization only once, and set to handle all the `policies` that matches the condition of a system payload (a system check-in to insights). So I don't think is short-lived and used by a single thread.

I still don't know much about the inner workings of [hawkular-alerts](https://github.com/hawkular/hawkular-alerts) (the project which this one forked from) or the engine itself. I'll try to seek better answers for this.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/notification/PoliciesAction.java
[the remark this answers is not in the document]
gwenneg: Oh I was assuming that there could be different `Event` objects references for a single event. There's no problem if the reference is always the same so I'm resolving this.

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
gwenneg: It looks like the bean should be `@ApplicationScoped` instead of `@Dependent`. It's not related to this PR though so I'm resolving this discussion.

gwenneg (approved):
I'm not an expert of policies-engine yet, but this LGTM

gwenneg (approved):

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
josejulio: @spinder Do you manage the grafana boards? This would need an update on ci/qa after we merge this one. And there is a new one below (`engine.actions.notifications.aggregated`)

PR #150, opened 2021-05-05
add dependabot

bennetelli:


bennetelli:
Somehow there is dependabot configured already, but I can't see where it comes from. 

I would prefer the file based configuration instead of a "hidden" configuration somehow in the github settings.

gwenneg:
I think it's configured in the repository settings. but I can't confirm since I don't have the permissions to access it.

bennetelli:
Me neither. Maybe @theute or @pilhuhn can take a loot at the repository configuration so we can remove dependatbot from there and have it in the repo only.

theute:
Disabled

gwenneg (approved):

PR #151, opened 2021-05-05
POL-434 add cloudwatch logging

bennetelli:


pilhuhn (approved):

external/src/main/resources/application.properties
```diff
@@ -100,3 +100,11 @@ quarkus.jaeger.sampler-param=1
 # Depending on the location, the default port is 14268
 #quarkus.jaeger.endpoint=http://jaeger-c.apps-crc.testing/api/traces
 
+quarkus.log.level=INFO
+quarkus.log.cloudwatch.enabled=true
```
theute, line 104: Shouldn't it be false by default ?

bennetelli: Indeed. Heiko and me also talked about it.

PR #166, opened 2021-05-20
POL-497 Override the Maven version used by Quarkus

gwenneg:
@aloubyansky Could you please confirm this is correct?


aloubyansky:
This looks good in the BOM. However, if you want to also update the Maven version Quarkus uses in dev mode, you should also override the Maven plugin dependencies.

gwenneg:
Quarkus was bumped to `1.13.6.Final` which includes a Maven upgrade (see [1.13.5.Final blog post](https://quarkus.io/blog/quarkus-1-13-5-final-released/)) so we don't need to override the Maven version anymore.

PR #183, opened 2021-06-10
POL-502 Add silenced logs

bennetelli:


bennetelli:
I don't have logs here. So maybe we should wait until the errors on OpenShift ci are fixed and the pod is running again so I can take a look at the access logs here. Maybe they are different than the others.

gwenneg (approved):
LGTM

PR #185, opened 2021-06-11
Print whitelisted env vars and conf values at application startup

gwenneg:
⚠️ Temp PR, will be reverted after we're done fixing the issue.

This PR will print all environment variables keys at application startup.

Values will also be printed but only if they start with a String that matches one of the elements of `PREFIX_WHITELIST`.

### DIFF
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
@@ -80,8 +80,8 @@ public class IspnAlertsServiceImpl implements AlertsService {
     boolean saveThinAlerts = false;
 
     public void init() {
-        alertsLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.alerts-lifespan", Long.class);
         eventLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.events-lifespan", Long.class);
+        alertsLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.alerts-lifespan", Long.class);
         saveThinAlerts = ConfigProvider.getConfig().getValue("engine.backend.ispn.alerts-thin", Boolean.class);
         backend = IspnCacheManager.getCacheManager().getCache("backend");
         if (backend == null) {
--- a/external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
@@ -3,6 +3,7 @@ package com.redhat.cloud.policies.engine;
 import com.redhat.cloud.policies.engine.actions.QuarkusActionPluginRegister;
 import io.quarkus.runtime.ShutdownEvent;
 import io.quarkus.runtime.StartupEvent;
+import org.eclipse.microprofile.config.ConfigProvider;
 import org.hawkular.alerts.AlertsStandalone;
 import org.hawkular.alerts.api.services.ActionsService;
 import org.hawkular.alerts.api.services.AlertsService;
@@ -16,6 +17,10 @@ import javax.enterprise.event.Observes;
 import javax.enterprise.inject.Produces;
 import javax.inject.Inject;
 import javax.inject.Singleton;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 
 import static org.hawkular.alerts.api.util.Util.isEmpty;
 
@@ -49,6 +54,7 @@ public class AlertStarter {
     }
 
     void startApp(@Observes StartupEvent startup) {
+        printEnvVars();
         String commit = System.getenv(BUILD_COMMIT_ENV_NAME);
         if(!isEmpty(commit)) {
             LOGGER.infof("Starting Policies Engine, build: %s", commit);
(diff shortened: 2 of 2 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
```diff
@@ -58,6 +62,29 @@ void startApp(@Observes StartupEvent startup) {
...
+            String value = isWhitelisted(var.getKey()) ? var.getValue() : "NOT_WHITELISTED";
+            LOGGER.info(var.getKey() + "=" + value);
+        }
+        LOGGER.info("=== ENVIRONMENT VARIABLES - END ===");
+    }
+
+    private static final List<String> PREFIX_WHITELIST = List.of(
+            "QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME",
+            "ENGINE"
+    );
```
gwenneg, line 77: This must contain all env vars prefixes we want to print.

gwenneg:
@theute @pilhuhn @bennetelli Any suggestions for the whitelist?

pilhuhn:
HOSTNAME was a topic.

gwenneg:
I'm not sure HOSTNAME is directly available as an env var. Our OpenShift configuration maps `${HOSTNAME}` to `QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME` which is why I added `QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME` to the whitelist.

external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
```diff
@@ -49,6 +52,7 @@ public ActionsService getActionsService() {
     }
 
     void startApp(@Observes StartupEvent startup) {
+        printEnvVars();
```
pilhuhn, line 55: Are we sure this is early enough? I don't recall if config resolution happens earlier or not.
And then it may be no issue on engine, as the values are not injected, but obtained by running code.

gwenneg: The current prod failure happens after that so it should be ok.

theute:
> I'm not sure HOSTNAME is directly available as an env var. Our OpenShift configuration maps `${HOSTNAME}` to `QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME` which is why I added `QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME` to the whitelist.

Yes but on a specific cluster HOSTNAME doesn't seem to resolve (QUARKUS_LOG_CLOUDWATCH_LOG_STREAM_NAME is empty) while it should be available, could you add HOSTNAME ?

bennetelli:
lgtm

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
```diff
@@ -80,8 +80,8 @@
     boolean saveThinAlerts = false;
 
     public void init() {
-        alertsLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.alerts-lifespan", Long.class);
         eventLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.events-lifespan", Long.class);
+        alertsLifespanInHours = ConfigProvider.getConfig().getValue("engine.backend.ispn.alerts-lifespan", Long.class);
```
josejulio, line 84: I assume we want to check if `engine.backend.ispn.events-lifespan` is the next fail?

external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
```diff
@@ -58,6 +62,30 @@ void startApp(@Observes StartupEvent startup) {
         initialize();
     }
 
+    private void printEnvVars() {
+        LOGGER.info("=== ENVIRONMENT VARIABLES - START ===");
```
josejulio, line 66: Since `ConfigProvider.getConfig().getValue` is the one failing, is there any way to print all the config values? (or ones prefixed with `engine` or so?)

pilhuhn: Yes, that is doable. I've sort of done similar in the ClowderConfigSource: https://github.com/RedHatInsights/clowder-quarkus-config-source/blob/7554c6108e2520a8e307955adabf9cd72d27058f/src/main/java/com/redhat/cloud/common/clowder/configsource/ClowderConfigSourceFactory.java#L39 my require a `ConfigSourceFactory` though (not sure)

engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnAlertsServiceImpl.java
gwenneg: Yes, we want to check if the issue affects all configuration values or only `engine.backend.ispn.alerts-lifespan`.

external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
gwenneg: This should work (with proper whitelist filtering):
```
for (String key : ConfigProvider.getConfig().getPropertyNames()) {
    LOGGER.info(key + "=" + ConfigProvider.getConfig().getValue(key, String.class));
}
```

gwenneg:
Quarkus conf values are now printed for keys starting with `engine.`.

PR #186, opened 2021-06-14
Fix config loading race condition

gwenneg:
This could be the solution to the prod config loading issue.

I was able to reproduce the issue locally. It only happens when the app is started with a `java -jar` command.

With the changes from this PR, I was no longer able to reproduce it.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
@@ -3,7 +3,6 @@ package com.redhat.cloud.policies.engine;
 import com.redhat.cloud.policies.engine.actions.QuarkusActionPluginRegister;
 import io.quarkus.runtime.ShutdownEvent;
 import io.quarkus.runtime.StartupEvent;
-import org.eclipse.microprofile.config.ConfigProvider;
 import org.hawkular.alerts.AlertsStandalone;
 import org.hawkular.alerts.api.services.ActionsService;
 import org.hawkular.alerts.api.services.AlertsService;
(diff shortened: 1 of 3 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/AlertStarter.java
```diff
@@ -64,56 +58,21 @@ void startApp(@Observes StartupEvent startup) {
...
-                .whenComplete((empty, error) -> {
-                    if(error != null) {
-                        LOGGER.error("Engine start failed", error);
-                        alerts.stop();
-                    } else {
-                        pluginRegister.init();
-                        ((StatusServiceImpl) statusService).setStarted(true);
-                        LOGGER.info("Started Policies Engine");
-                    }
-        }).join();
```
gwenneg, line 117: This was nothing more than a sequential execution using a thread from `ForkJoinPool`.

external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
```diff
@@ -153,36 +153,34 @@ public AlertsStandalone() {
...
-
-            partitionManager.init();
-            alertsContext.init();
-            dataDrivenGroupCacheManager.init();
-            actionsCacheManager.init();
-            publishCacheManager.init();
-            extensions.init();
-            engine.initServices();
-        });
+    public void init() {
```
gwenneg, line 156: This method is only called by `AlertStarter#initialize` which doesn't seem async execution. It's ok to drop the `CompletableFuture<Void>` return type.

gwenneg (commented):
Review with `Hide whitespace changes` checked for an easier review.

pilhuhn:
I can't comment if this fixes the code or not, but I certainly like the change, as it simplifies the code.

josejulio:
> I can't comment if this fixes the code or not, but I certainly like the change, as it simplifies the code.

ditto

josejulio (approved):
lgtm

PR #189, opened 2021-06-22
POL-532 Process Kafka messages in a non-blocking way

gwenneg:
This may fix the `VertxException` we have on prod.

Even if it does not fix the exception, it is still an improvement: Kafka messages are now processed in a non-blocking way.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
@@ -2,6 +2,7 @@ package com.redhat.cloud.policies.engine.process;
 
 import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
+import io.smallrye.mutiny.Uni;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import org.eclipse.microprofile.config.inject.ConfigProperty;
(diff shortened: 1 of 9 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
```diff
@@ -184,15 +184,19 @@
             List<Event> eventList = new ArrayList<>(1);
             eventList.add(event);
             if (storeEvents) {
-                alertsService.addEvents(eventList);
+                return alertsService.addEvents(eventList)
+                        .replaceWith(ack(input));
             } else {
-                alertsService.sendEvents(eventList);
+                return alertsService.sendEvents(eventList)
+                        .replaceWith(ack(input));
```
gwenneg, line 191: This is the place where the processing was changed from blocking to non-blocking.

josejulio: If there is an error in the `sendEvents`, it won't `ack` the input?
or it `ack`s it regardless of the outcome?

gwenneg: An exception will be thrown and `ack` will never be called. It should behave like the old code. Maybe the exception thrown will be different though, Mutiny may wrap the original exception into a Mutiny exception. But I don't think it will change anything in terms of application behavior.

josejulio (approved):

PR #196, opened 2021-06-28
Process events from a worker thread

gwenneg:
This will hopefully fix the `io.vertx.core.VertxException: Thread blocked` we have on prod.


gwenneg:
The exception is usually thrown on long-running pods when the filesystem usage is high (around 40M in the OpenShift metrics) so we will have to wait a bit before we can determine if the issue is fixed or not.

gwenneg:
We should also verify if this can be reverted after the policies-history integration.

pilhuhn (approved):
lgtm

PR #197, opened 2021-06-28
POL-450 Persist events and alerts with PostgreSQL

gwenneg:
⚠️ Big PR update in progress...


gwenneg:
The switch between the Infinispan and the policies-history implementations is done using the [@IfBuildProperty](https://quarkus.io/guides/cdi-reference#enabling-beans-for-quarkus-build-properties) feature from ArC (Quarkus):

```
@ApplicationScoped
@IfBuildProperty(name = POLICIES_HISTORY_ENABLED_CONF_KEY, stringValue = "true")
public class PoliciesHistoryActionsService implements ActionsService {

    public static final String POLICIES_HISTORY_ENABLED_CONF_KEY = "policies-history.enabled";

[...]
```

```
public class IspnActionsServiceImplProducer {

    @Singleton
    @IfBuildProperty(name = POLICIES_HISTORY_ENABLED_CONF_KEY, stringValue = "false")
    public ActionsService getActionsService() {
        return new IspnActionsServiceImpl();
    }
}
```

```
@Singleton
public class AlertsStandalone {

    @Inject
    ActionsService actionsService;

[...]
```

external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
```diff
@@ -65,14 +66,14 @@
...
     private IspnDefinitionsServiceImpl ispnDefinitions;
     private StatusServiceImpl status;
     private PartitionManagerImpl partitionManager;
     private PublishCacheManager publishCacheManager;
     private IspnAdminService adminService;
 
-    public AlertsStandalone() {
+    @PostConstruct
+    void postConstruct() {
         log.info("Policies Engine uses Infinispan backend");
```
pilhuhn, line 77: Is that still true? See L107

gwenneg: Depending on the `policies-history.alerts-service.enabled` conf value, it can be completely true or partly true 😃

gwenneg:
We won't do that for now. Replaced with #207.

PR #199, opened 2021-07-05
Lower the number of objects stored in the heap old gen

gwenneg:
I use JFR / MissionControl for my `policies-history` work to check the impact of code changes on the prod memory leak (which is really easy to reproduce locally BTW). Heap old gen analysis showed a lot of `ArrayList`/`HashSet` resizes and also cache `read` operations after `put` operations while we don't use the returned values. This PR contains short-term optimizations that will lower the number of objects stored in the old gen. It only mitigates some of the consequences of the leak but does not fix the cause at all. Don't expect a fantastic result, but there will be a slight improvement.

See [this article](https://infinispan.org/blog/2012/09/04/speeding-up-cache-calls-with) about the reason why we should use `Flag.IGNORE_RETURN_VALUES`.

### DIFF
--- a/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
+++ b/engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
@@ -59,6 +59,8 @@ import org.infinispan.query.dsl.FilterConditionContext;
 import org.infinispan.query.dsl.QueryBuilder;
 import org.infinispan.query.dsl.QueryFactory;
 
+import static org.infinispan.context.Flag.IGNORE_RETURN_VALUES;
+
 /**
  * @author Jay Shaughnessy
  * @author Lucas Ponce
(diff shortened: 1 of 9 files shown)


engine/src/main/java/org/hawkular/alerts/engine/impl/ispn/IspnDefinitionsServiceImpl.java
```diff
@@ -138,7 +140,7 @@ public void addActionDefinition(String tenantId, ActionDefinition actionDefiniti
         if (found != null) {
             throw new FoundException(pk);
         }
-        backend.put(pk(actionDefinition), new IspnActionDefinition(actionDefinition));
+        backend.getAdvancedCache().withFlags(IGNORE_RETURN_VALUES).put(pk(actionDefinition), new IspnActionDefinition(actionDefinition));
```
pilhuhn, line 143: Would it make sense to turn `backend` into a new class `SuperCache` (that inherits from `Cache`) which then gets a method `putNR()`, which internally supplies the flag? This way you would not need the `.getAC().withFlags()` all over the place

gwenneg: If we wanted to keep using `backend`, yes definitely. Now, with policies-history coming soon (hopefully!) I don't know if that's worth it.

gwenneg: @pilhuhn Can we merge the current version or would you prefer the improved one?

pilhuhn: Whatever you find right :)

gwenneg: Let's merge this then, I'd rather stay focused on the main part 😃

josejulio (approved):

PR #203, opened 2021-07-05
POL-542 Add clowdapp for engine

lahavyuv86:


theute:
Don't we need all the environment variables as we find in e2e-deploy ?
IIUC the plan is to use Clowder for all environments so we'll ultimately need all those variables, or am I missing something ?

lahavyuv86:
I can add them NP, just wanted to know if all is needed or just some of them

theute:
I think only a few are needed for ephemeral, but all will be needed once we start using them for all environments (this is true for the 4 pods)

lahavyuv86:
@theute The env vars were added, please review again

theute:
Worth a try :)

PR #207, opened 2021-07-13
POL-450 Store alerts history entries into PostgreSQL

gwenneg:
@pilhuhn Here's a new draft following our policies history meeting from yesterday:

- A `PoliciesHistoryEntry` instance is created and persisted into the database everytime a new `Alert` is added into the Infinispan `backend` cache.
- There isn't any change regarding the data persisted into the Infinispan caches.
- There's almost nothing new in Hawkular (only one service interface and a method call), most of the code is in the `external` module.
- The persisted data will be consumed by `policies-ui-backend` directly from the database.
- The new service is disabled by default for now.
- If an exception is thrown while persisting an alert history data, the exception will be logged at `ERROR` level and Hawkular will keep running normally.
- The new service logs a message similar to this at application startup `2021-07-13 10:33:43,310 INFO  [com.red.clo.pol.eng.his.AlertsHistoryServiceImpl] (main) Policies history is disabled. The history won't be persisted into the database.` (or enabled, depending on the configuration)
- The persisted data look like this:
```
policies_ui_backend=# select * from alerts_history;
                  id                  |     ctime     |               host_id                | host_name |                     policy_id                     |    tenant_id
--------------------------------------+---------------+--------------------------------------+-----------+---------------------------------------------------+------------------
 f92d504d-984d-4fa6-8232-75d3ac0792f4 | 1626164089237 | ba11a21a-8b22-431b-9b4b-b06006472d54 | VM        | arch-triggerd09081b2-f8da-4127-b1ed-87e04e10b8be  | integration-test
 3d4fcc1c-fd91-4c98-b5dd-62d0fbb869d8 | 1626164089260 | ba11a21a-8b22-431b-9b4b-b06006472d54 | VM        | arch-triggerd09081b2-f8da-4127-b1ed-87e04e10b8be2 | integration-test
 5520b3b2-58ab-4102-86e6-69348b207ca3 | 1626165444189 | ba11a21a-8b22-431b-9b4b-b06006472d54 | VM        | arch-triggerb60884e7-980a-467b-b60e-c6721d5df530  | integration-test
(3 rows)
```

⚠️ I'm not done with this PR, still working on some details (mostly tests).

⚠️ Consuming the data directly from `policies-ui-backend` with JPA queries is simple so I really like this idea, but it comes with a minor drawback: we have to choose which project is responsible for maintaining the `alerts_history` table schema. Since ui-backend is the current "owner" of the `policies_ui_backend` database, I think it should be responsible for the `alerts_history` table, so I added in this PR a Flyway script in test mode only.

Does that look better than the previous version? 😃 

cc @theute


external/src/test/resources/migration/V1.0.0__POL-450_alerts_history.sql
```diff
@@ -0,0 +1,13 @@
...
+-- This project uses the `policies_ui_backend` database but its schema is actually managed by the policies-ui-backend project.
+-- We need that script to run standalone policies-engine tests.
+
+create table alerts_history (
+    id uuid not null,
+    tenant_id varchar(255) not null,
+    policy_id varchar(255) not null,
+    ctime bigint not null,
+    host_id varchar(255),
+    host_name varchar(255),
```
gwenneg, line 11: These `varchar` max lengths probably need to be adjusted.

external/src/main/java/com/redhat/cloud/policies/engine/history/AlertsHistoryEntry.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import java.util.Objects;
+import java.util.UUID;
+
+@Entity
+@Table(name = "alerts_history")
+public class AlertsHistoryEntry {
```
gwenneg, line 12: Policies-ui-backend will need a similar class when reading from the database. One duplicated class isn't so bad but it would be even better if we could avoid duplicated code.

theute:
Sounds good to me, just some non-blocking comments:
- We'll need to find a way to stop storing into ISPN
- We'll need a way to cleanup events older than 14 days
- For data migration, since we store in both for now, we could:
1) Use both storage
2) In 14 days (or whenever we are ready) change the read from the old to the new storage (we don't have to read from both)
3) Stop storing in ISPN
(we could even do: 4) delete all ISPN data and 5) recreate the triggers to get rid of all ISPN history)

external/src/test/resources/migration/V1.0.0__POL-450_alerts_history.sql
pilhuhn: Shouldn't the Id's be uuid? So something like 32 chars?

gwenneg: Yes probably, but these values are currently stored as `String` and not `UUID` into Infinispan so I chose the safest path with fields capacities purposedly large (maybe a bit too much). We can easily alter the schema and move from `varchar(255)` to `uuid` (or a small `varchar`) later. I thought about "dumping" prod data after one or two weeks and check the max length of each field.

external/src/main/java/com/redhat/cloud/policies/engine/history/PoliciesHistoryEntry.java
```diff
@@ -0,0 +1,97 @@
...
+
+@Entity
+@Table(name = "policies_history")
+public class PoliciesHistoryEntry {
+
+    @Id
+    @GeneratedValue
+    private UUID id;
+
+    private String tenantId;
```
josejulio, line 18: Same comment as in the `policies-ui-backend`, should we call this `accountId` ? `tenant` is a term used by hawkular-alerts, akin to say `triggerId` instead of `policyId`.

josejulio: Something else is that in `policies-ui-backend`, the fields are annotated with e.g. `@Column(name = "tenant_id")`, but not on `policies-engine`

external/src/main/resources/application.properties
```diff
@@ -130,3 +130,24 @@ quarkus.log.cloudwatch.access-key-secret=placeholder
...
+
+# The following database is used to store the policies history data which is then used from the policies-ui-backend project.
+# The database schema is managed by policies-ui-backend.
+quarkus.datasource.db-kind=postgresql
+quarkus.datasource.jdbc.url=jdbc:postgresql://127.0.0.1:5432/postgres
+quarkus.datasource.username=postgres
+quarkus.datasource.password=postgres
+
+# Transforms camel case entities attributes names into snake case.
+quarkus.hibernate-orm.physical-naming-strategy=com.redhat.cloud.policies.engine.history.SnakeCasePhysicalNamingStrategy
```
josejulio, line 150: Could this be an annotation? Or maybe these don't work here and that's the reason of the lack of `@Column` annotation in fields?

gwenneg: `SnakeCasePhysicalNamingStrategy` is my preferred way of converting Java fields names into DB fields names because it is global, but I didn't use it in `policies-ui-backend` because that project already had some persistence code based on Panache, so I used `@Column` annotations instead to avoid introducing a regression. On the other hand, `policies-engine` didn't have any database-related code so I added `SnakeCasePhysicalNamingStrategy`.

external/src/main/java/com/redhat/cloud/policies/engine/history/PoliciesHistoryEntry.java
gwenneg: > Same comment as in the `policies-ui-backend`, should we call this `accountId` ? `tenant` is a term used by hawkular-alerts, akin to say `triggerId` instead of `policyId`.

Same kind of reply than in policies-ui-backend: `tenantId` is everywhere there 😃

gwenneg: > Something else is that in `policies-ui-backend`, the fields are annotated with e.g. `@Column(name = "tenant_id")`, but not on `policies-engine`

See [this](https://github.com/RedHatInsights/policies-engine/pull/207#discussion_r675075264).

josejulio (approved):
lgtm

PR #225, opened 2021-08-19
Prevent non-application path regressions

gwenneg:
Same as https://github.com/RedHatInsights/notifications-backend/pull/501.


external/src/main/resources/application.properties
```diff
@@ -146,3 +146,7 @@ quarkus.hibernate-orm.physical-naming-strategy=com.redhat.cloud.policies.engine.
 
 # Uncomment to log Hibernate SQL statements
 #quarkus.hibernate-orm.log.sql=true
+
+# Quarkus since 1.11 redirects non-apps to /q/. We need to prevent this
+quarkus.http.redirect-to-non-application-root-path=false
+quarkus.http.non-application-root-path=/
```
gwenneg, line 152: I need to determine the impacts of this change. The PR will be a draft until then.

PR #234, opened 2021-09-21
NOTIF-309 Send rh-message-id Kafka header

gwenneg:
Same as https://github.com/RedHatInsights/notifications-gw/pull/103.


pilhuhn (approved):

external/src/main/java/com/redhat/cloud/policies/engine/actions/plugins/NotificationActionPluginListener.java
```diff
@@ -50,12 +56,14 @@
...
     public static final String EVENT_TYPE_NAME = "policy-triggered";
+    public static final String WEBHOOK_CHANNEL = "webhook";
+    public static final String MESSAGE_ID_HEADER = "rh-message-id";
 
     private static final Logger log = Logger.getLogger(NotificationActionPluginListener.class.getName());
     private final ConcurrentSkipListMap<String, PoliciesAction> notifyBuffer = new ConcurrentSkipListMap<>();
 
     @Inject
-    @Channel("webhook")
+    @Channel(WEBHOOK_CHANNEL)
```
pilhuhn, line 66: Just a side note: I think we should rename this to something like 'notifications' or 'to-notifications' or 'outgoing' to make it clearer that this is for general notification processing.
Not in this PR of course :)

PR #276, opened 2021-11-29
POL-561 Stop persisting triggers history in Infinispan

gwenneg:
This PR disables the triggers lifecycle which means that when a trigger is fired, it will no longer:
- be loaded from Infinispan
- have its lifecycle updated
- be saved into Infinispan

ℹ️ This is tested with IQE tests on ephemeral and they all passed.

### DIFF
--- a/external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
+++ b/external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
@@ -3,6 +3,7 @@ package org.hawkular.alerts;
 import com.redhat.cloud.policies.engine.actions.QuarkusActionPluginRegister;
 import io.quarkus.runtime.LaunchMode;
 import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
 import org.hawkular.alerts.api.services.ActionsService;
 import org.hawkular.alerts.api.services.PoliciesHistoryService;
 import org.hawkular.alerts.api.services.AlertsService;
@@ -55,6 +56,10 @@ public class AlertsStandalone {
 
     @Inject
     PoliciesHistoryService policiesHistoryService;
+
+    // The triggers' lifecycle is no longer used by policies-ui-backend. It has been replaced with a Postgres table.
+    @ConfigProperty(name = "engine.triggers-lifecycle.disabled", defaultValue = "false")
+    boolean triggersLifecycleDisabled;
 
     //    @ConfigProperty(name = "engine.backend.ispn.reindex", defaultValue = "false")
     private boolean ispnReindex;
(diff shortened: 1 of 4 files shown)


theute:
SETM (Sounds Exciting To Me)

external/src/test/groovy/org/hawkular/alerts/rest/LifecycleITest.groovy
```diff
@@ -325,8 +326,9 @@ class LifecycleITest extends AbstractQuarkusITestBase {
     }
 
     @Test
+    @Disabled
```
bennetelli, line 329: should we delete the test instead of disabling it?

gwenneg: I'd rather keep it here for now. I disabled the triggers history Infinispan persistence in Hawkular using a property. Nothing got deleted or changed for now. This test would still be relevant if we decided to change the property value for some reason.

bennetelli: I don't like to keep unused lines of code because I think this code maybe never won't be used in the future and people will still have to read it and wonder why this is disabled.

If you would like to keep it, putting a reason @Disabled("disabled because of xyz") would be great so everyone knows why it's disabled.

gwenneg: Good point, thanks. I added the reason.

bennetelli: Great. Thanks for adding it :)

external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
```diff
@@ -56,6 +57,10 @@
     @Inject
     PoliciesHistoryService policiesHistoryService;
 
+    // The triggers' lifecycle is no longer used by policies-ui-backend. It has been replaced with a Postgres table.
+    @ConfigProperty(name = "engine.triggers-lifecycle.disabled", defaultValue = "false")
+    boolean triggersLifecycleDisabled;
+
     //    @ConfigProperty(name = "engine.backend.ispn.reindex", defaultValue = "false")
```
bennetelli, line 64: not related to your PR, but why is this uncommented?

bennetelli (approved):

external/src/main/java/org/hawkular/alerts/AlertsStandalone.java
gwenneg: I have no idea.

pilhuhn: Git blame on that line leads to https://github.com/RedHatInsights/policies-engine/commit/b2ec21832890ff90787ef3b4fcb525c44d621f33

gwenneg:
I pushed to trigger an ephemeral deployment, the code didn't change.

gwenneg:
Looks good on ephemeral.

PR #286, opened 2021-12-06
Disable actions store

gwenneg:
This PR may or may not have a significant impact on the stage/prod memory leak. 😄 

I am able to reproduce a memory leak on my machine using JMeter and a large number of incoming Kafka messages on the `platform.inventory.events` topic. As a consequence, I had the occasion to analyze a few heap dumps and to test many code changes.

Here's what I suspect. With the current `master` code, the garbage collector is unable to release the memory allocated to an `Alert` because of the `Action#event` field which holds a reference to the `Alert`. The `Action` is stored in a volatile Infinispan cache for a duration corresponding to the `Alert` lifespan configured in `application.properties` (currently 24h). After that delay, the `Alert` becomes eligible to garbage collection.

I don't think we need to persist even temporarily the actions in Infinispan. Our `ActionListener` immediately transforms each action into a Kafka message sent on `platform.notifications.ingress`, that should be enough for Policies to work. This is why I'd like to deploy this PR on stage and observe the memory afterwards.

If my asumptions are wrong then it will be super easy to revert the code.

### DIFF
--- a/external/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
+++ b/external/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
@@ -11,6 +11,7 @@ import org.hawkular.alerts.api.model.trigger.Trigger
 import org.hawkular.alerts.api.model.trigger.TriggerAction
 import org.hawkular.alerts.log.MsgLogger
 import org.hawkular.alerts.log.MsgLogging
+import org.junit.jupiter.api.Disabled
 import org.junit.jupiter.api.Test
 
 import static org.hawkular.alerts.api.model.event.Alert.Status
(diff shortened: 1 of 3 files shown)


gwenneg:
One of the tests failed, I'm having a look at it.

external/src/test/groovy/org/hawkular/alerts/rest/ActionsITest.groovy
```diff
@@ -283,6 +284,7 @@ class ActionsITest extends AbstractQuarkusITestBase {
     }
 
     @Test
+    @Disabled("The actions store is no longer used by the engine. This is still an experimental change so let's keep this test around for a while.")
```
gwenneg, line 287: I disabled this test and the other one below because they won't succeed unless the actions are persisted in Infinispan.

theute:
Works for me, I'll be happy to do some tests in stage once deployed

gwenneg:
Some IQE tests are failing now. I'm having a look.

gwenneg:
Looks unrelated at first glance (403). I'm running the IQE tests again and watching logs on ephemeral.

pilhuhn (approved):

PR #359, opened 2022-04-27
Log which Garbage Collector is used

bennetelli:


gwenneg:
The answer is [G1](https://github.com/RedHatInsights/policies-engine/blob/b98a952582c7278a1ee0dff8e81f2c39df815e12/.rhcicd/clowdapp.yaml#L98).

bennetelli:
> The answer is [G1](https://github.com/RedHatInsights/policies-engine/blob/b98a952582c7278a1ee0dff8e81f2c39df815e12/.rhcicd/clowdapp.yaml#L98).

huh. You already worked on it in the past. Then we don't need it :)

PR #363, opened 2022-05-02
POL-649 Introduce LightweightEngine

gwenneg:
This PR is the first step of the removal of Infinispan, Drools and most of Hawkular from `policies-engine`. This step is only focused on making the Kafka messages processing simpler.

:information_source: If the `lightweight-engine.enabled` config key is set to `false` (which will be the default value), this PR does not have any impact on prod and can be merged safely.

Now, let's talk about what's changed :smiley: 

:arrow_forward: **BEFORE**

before

:arrow_forward: **AFTER**

after

:arrow_forward: **Important details**

- All triggers are loaded in memory (in a `HashMap`). This should be equivalent to what we had before since the triggers were loaded in memory through Drools. However, we will have to monitor the heap for a while to make sure everything's fine.
- The old logic based on a `TimerTask` (scheduled batch events processing) is gone. Now, whenever a Kafka message is received on the `platform.inventory.events` topic, the tenant's triggers conditions are immediately tested and if they are satisfied, a Kafka message is immediately sent to the `platform.notifications.ingress` topic.
- Hawkular includes features to limit alerts spam. For example, it can enforce a minimum delay between two events and discard events that do not comply with that policy. This is gone as well. The new engine will process any incoming message and transform it into an outgoing notification if one or several triggers of the tenant are fired.

:arrow_forward: **Next step**

Stop using Infinispan to store triggers and replace it with Postgres.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
@@ -102,7 +102,9 @@ public class LightweightEngineImpl implements LightweightEngine {
         Map<String, FullTrigger> tenantTriggers = triggersByTenant.get(tenantId);
         if (tenantTriggers != null) {
             found = tenantTriggers.remove(triggerId) != null;
-            LOGGER.debugf("Trigger removed");
+            if (found) {
+                LOGGER.debugf("Trigger removed");
+            }
         }
         if (!found) {
             LOGGER.debugf("Trigger to remove not found");
--- a/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
@@ -1,0 +1,265 @@
+package com.redhat.cloud.policies.engine;
+
+import com.redhat.cloud.policies.engine.actions.plugins.notification.PoliciesAction;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.reactive.messaging.Channel;
+import org.eclipse.microprofile.reactive.messaging.Emitter;
+import org.eclipse.microprofile.reactive.messaging.OnOverflow;
+import org.hawkular.alerts.api.model.condition.Condition;
+import org.hawkular.alerts.api.model.condition.ConditionEval;
+import org.hawkular.alerts.api.model.condition.EventCondition;
+import org.hawkular.alerts.api.model.condition.EventConditionEval;
+import org.hawkular.alerts.api.model.dampening.Dampening;
+import org.hawkular.alerts.api.model.event.Alert;
+import org.hawkular.alerts.api.model.event.Event;
+import org.hawkular.alerts.api.model.trigger.FullTrigger;
+import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.services.LightweightEngine;
+import org.hawkular.alerts.api.services.PoliciesHistoryService;
+import org.jboss.logging.Logger;
+
+import javax.annotation.PostConstruct;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.io.IOException;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static com.redhat.cloud.policies.engine.actions.plugins.NotificationActionPluginListener.WEBHOOK_CHANNEL;
+import static com.redhat.cloud.policies.engine.actions.plugins.NotificationActionPluginListener.buildMessageWithId;
+import static com.redhat.cloud.policies.engine.actions.plugins.NotificationActionPluginListener.serializeAction;
+import static java.time.ZoneOffset.UTC;
+import static org.eclipse.microprofile.reactive.messaging.OnOverflow.Strategy.UNBOUNDED_BUFFER;
+
+@ApplicationScoped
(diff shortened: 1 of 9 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
```diff
@@ -184,20 +191,25 @@ public Uni<Void> processAsync(Message<String> input) {
...
-                        .replaceWith(ack(input))
-                        .runSubscriptionOn(Infrastructure.getDefaultWorkerPool());
+        if (lightweightEngineEnabled) {
+            lightweightEngine.process(event);
+            return Uni.createFrom().voidItem();
+        } else {
+            try {
+                List<Event> eventList = new ArrayList<>(1);
+                eventList.add(event);
+                if (storeEvents) {
```
josejulio, line 201: Isn't this something we could already remove? 
IIRC it's part of a previous migration to store in psql? or is something else?

gwenneg: That branch of the `if` is already unused so it could be removed, yes.

However, I don't plan on removing anything yet. There will be a huge cleaning PR in the future, after we're done testing the new engine on prod.

external/src/main/resources/application.properties
```diff
@@ -149,3 +151,8 @@ quarkus.http.non-application-root-path=/
 
 engine.triggers-lifecycle.disabled=true
 engine.actions-store.disabled=true
+
+# Enables LightweightEngine which is an engine implementation that does not rely on Infinispan or Drools.
+lightweight-engine.enabled=true
```
gwenneg, line 156: This is enabled to show that test are passing with the new implementation. I will disable it before we merge the PR though.

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    @Override
+    public void removeTrigger(String tenantId, String triggerId) {
+        LOGGER.debugf("Removing trigger %s/%s", tenantId, triggerId);
+        boolean found = false;
+        Map<String, FullTrigger> tenantTriggers = triggersByTenant.get(tenantId);
+        if (tenantTriggers != null) {
+            found = tenantTriggers.remove(triggerId) != null;
+            LOGGER.debugf("Trigger removed");
```
josejulio, line 105: nit, bit misleading
```suggestion
            if (found) {
                LOGGER.debugf("Trigger removed");
            }
```

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
```diff
@@ -0,0 +1,265 @@
...
+    public void process(Event event) {
+        LOGGER.debugf("Processing event %s", event);
+        Set<Alert> alerts = fireTriggers(event);
+        if (alerts.isEmpty()) {
+            LOGGER.debug("No alerts created from the event");
+        } else {
+            LOGGER.debugf("%d alerts created from the event", alerts.size());
+
+            // In the old implementation, the time comes from the Action constructor which relies on System.currentTimeMillis().
+            LocalDateTime nowUTC = LocalDateTime.ofInstant(Instant.ofEpochMilli(System.currentTimeMillis()), UTC);
```
josejulio, line 122: I guess it would be OK to use `LocalDateTime.now`, would that be a way different value? We can start with this currentTimeMillis anyhow.

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    /*
+     * Java equivalent of the old Event rule.
+     * See the ConditionMatch.drl file for more details about that rule.
+     */
+    private Optional<EventConditionEval> evaluateCondition(Trigger trigger, EventCondition eventCondition, Event event) {
+        if (trigger.getMode() == eventCondition.getTriggerMode() && trigger.getSource().equals(event.getDataSource())
+                && eventCondition.getDataId().equals(event.getDataId())) {
```
josejulio, line 248: What does it mean that the `EventCondition` and `Event` have different `mode`, `source` or `dataId` ? Is this something we should log?

josejulio (approved):
Muy bien!

Some minor comments/questions below

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
gwenneg: Good catch.

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
gwenneg: This Java code is a strict translation of the Drools rule. That old rule didn't log anything when these conditions were not met so I don't think we need to log anything.

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
gwenneg: We could use it, yes. I purposely changed as few code as possible in this PR to avoid introducing bugs and behavior differences though, so let's keep that change for another PR if that's fine with you :smiley:

gwenneg:
As previously announced, `LightweightEngine` is now disabled by default. It will be enabled on stage from the `app-interface` configuration.

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
josejulio: Sounds good!

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
josejulio: It's fine by me.

josejulio (approved):

PR #370, opened 2022-05-04
POL-649 Check trigger action before sending a notification

gwenneg:
Before this PR, the new engine sends notifications even if `notifications` hasn't been selected as a policy action in the frontend.

### DIFF
--- a/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
+++ b/external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
@@ -16,6 +16,7 @@ import org.hawkular.alerts.api.model.event.Alert;
 import org.hawkular.alerts.api.model.event.Event;
 import org.hawkular.alerts.api.model.trigger.FullTrigger;
 import org.hawkular.alerts.api.model.trigger.Trigger;
+import org.hawkular.alerts.api.model.trigger.TriggerAction;
 import org.hawkular.alerts.api.services.LightweightEngine;
 import org.hawkular.alerts.api.services.PoliciesHistoryService;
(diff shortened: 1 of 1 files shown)


external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
```diff
@@ -257,6 +260,18 @@ private Set<Alert> fireTriggers(Event event) {
         }
     }
 
+    private boolean shouldSendNotification(FullTrigger fullTrigger) {
+        if (fullTrigger.getTrigger().getActions() != null) {
+            for (TriggerAction triggerAction : fullTrigger.getTrigger().getActions()) {
+                if (triggerAction.getActionPlugin().equals("email") || triggerAction.getActionPlugin().equals("notification")) {
```
gwenneg, line 266: @josejulio I'm not about checking `notification`, it might not be necessary. WDYT?

josejulio: This looks good - Wondering now if we verify if the trigger/policy is enabled before processing it - i don't recall seeing that but unsure if it was somewhere else outside of the LightweightEngine.

gwenneg: Good question. I already checked with the logs that triggers are (re)loaded in `LightweightEngineImpl` when enabled. I need to check that they are removed when disabled.

gwenneg: Something's fishy with disabled triggers. I'll investigate that a bit later.

PR #375, opened 2022-05-05
POL-649 Introduce lightweight engine REST API and load triggers from DB

gwenneg:
Related to https://github.com/RedHatInsights/policies-ui-backend/pull/420.

This PR introduces the lightweight engine REST API and the triggers loading from the `policies-ui-backend` DB.

It changes a few things in the lightweight engine configuration:
| Old Config key | New config key | Default value | Description |
| - | - | - | - |
| lightweight-engine.enabled | lightweight-engine.kafka-processing.enabled | false | Enables the Kafka messages processing with the lightweight engine |
| | lightweight-engine.db-loading.enabled | false | Enables the lightweight engine REST API and the triggers loading from the DB |

ℹ️ As usual, all changes are disabled by default which makes this PR safe to merge.

⚠️ Check the `Hide whitespace` option for an easier review.


gwenneg:
ℹ️ I'm starting local tests with both `engine` and `ui-backend` PRs. They can be reviewed in the meantime.

api/src/main/java/org/hawkular/alerts/api/services/LightweightEngine.java
```diff
@@ -6,31 +6,59 @@
...
 
 /**
- * Simple engine implementation that does not rely on Infinispan or Drools for the Kafka messages processing. This is
- * the first step of a more important work aimed at making policies-engine simpler.
+ * Simple engine implementation that does not rely on Infinispan or Drools for the Kafka messages processing.
  */
 public interface LightweightEngine {
 
+    /**
+     * Loads all policies from the DB and converts them into full triggers that are stored in memory by the engine.
```
josejulio, line 18: I'm wondering, would it be possible as a next step to load the triggers on demand (i.e. when they are needed). 
I guess that would break the remaining link between the policies-engine and the policies-ui-backend.

gwenneg: We could load all the triggers of a tenant/account on demand. But we can't load triggers on demand individually because when we receive a message from a tenant, we have to test all his triggers.

josejulio: Yes, I was wondering the former. Loading all the triggers of a tenant on demand (to test them all).

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineImpl.java
```diff
@@ -229,61 +297,73 @@ private Set<Alert> fireTriggers(Event event) {
...
         }
     }
 
     private boolean shouldSendNotification(FullTrigger fullTrigger) {
-        if (fullTrigger.getTrigger().getActions() != null) {
-            for (TriggerAction triggerAction : fullTrigger.getTrigger().getActions()) {
-                if (triggerAction.getActionPlugin().equals("email") || triggerAction.getActionPlugin().equals("notification")) {
-                    return true;
+        if (lightweightEngineConfig.isRestApiEnabled()) {
+            return true;
```
josejulio, line 335: Why it returns true without checking the action plugins of the trigger? Is this done elsewhere?

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/PolicyToTriggerConverter.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class PolicyToTriggerConverter {
+
+    private static final String NO_LONGER_USED = "";
+
+    public static FullTrigger convert(Policy policy) {
+        FullTrigger fullTrigger = new FullTrigger();
+        fullTrigger.setTrigger(buildTrigger(policy));
+        fullTrigger.setConditions(List.of(buildCondition(fullTrigger.getTrigger(), policy.condition)));
+        return fullTrigger;
```
josejulio, line 18: I just noticed that this method is not loading the `policy.actions`.
So right now we could have:
- no actions
- notification action

In the future we are expecting more actions such as assigning a tag to the host. 

As i understand, right now this assumes all the policies should send a notification.

gwenneg: I forgot that a trigger could have no action in the UI 🤦‍♂️  I'll fix that and put the action check you mentioned in the other comment back in the code.

gwenneg: Actions should now be loaded from the DB and checked before sending a notification.

josejulio (approved):
Terrific!

gwenneg:
Self-review complete. It's merging time! 🎉

PR #377, opened 2022-05-12
POL-650 Add orgId to PoliciesAction

bennetelli:


josejulio:
> @josejulio not sure if I understand how that Event thing here works in detail, but it looks like someone can set the tenantId which could either be an accoundId or even an orgId. Is that correct?

I guess so, the `tenantId` comes from the hawkular world, It's stored in the infinispan IIRC, but with @gwenneg effort to simplify it, we will load it from the `policy` table. I guess we could use the `tenantId` as both, and decide what to use at loading time: https://github.com/RedHatInsights/policies-engine/pull/375/files#diff-c051bda5c3e1d666beabac8adce5724658594379c6643f39fe8caafd55516d80R27

external/src/main/java/com/redhat/cloud/policies/engine/LightweightEngineImpl.java
```diff
@@ -138,7 +144,11 @@ public void process(Event event) {
             // In the old implementation, the time comes from the Action constructor which relies on System.currentTimeMillis().
             LocalDateTime nowUTC = LocalDateTime.ofInstant(Instant.ofEpochMilli(System.currentTimeMillis()), UTC);
 
-            PoliciesAction policiesAction = new PoliciesAction();
+            PoliciesAction policiesAction = new PoliciesAction(useOrgId);
+
+            // TODO POL-650 is tenantId also the orgID? I guess it will become the ordId.
+            policiesAction.setOrgId(event.getTenantId());
```
josejulio, line 150: Maybe depending on the `useOrgId` value we would set it on the OrgId or the AccountId

bennetelli: thought about this as well. I did it this way to keep policies-engine compatible in every case. But maybe that's not needed. Not sure. Maybe @gwenneg has got an opinion about it?

bennetelli:
> > @gwenneg why do we have an Entity "policy" but no table for it?
> 
> The table is specified in the [policies-ui-backend](https://github.com/RedHatInsights/policies-ui-backend/tree/master/src/main/resources/db/migration)

aaah- I didn't think about that we could have a shared database. Thanks

engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
```diff
@@ -179,11 +183,11 @@ public void thresholdTest() {
         assertEquals(e.toString(), 1, e.getConditionSetSize());
         assertEquals("trigger-2", e.getTriggerId());
         assertTrue(e.isMatch());
-        assertTrue(!v.equals(e.getValue()));
+        assertFalse(v.equals(e.getValue()));
         v = e.getValue();
         assertTrue(e.toString(), v.equals(5.0D) || v.equals(10.0D));
         assertEquals(e.getCondition().toString(), "NumericData-01", e.getCondition().getDataId());
-        assertTrue(e.getCondition().getDataId().equals("NumericData-01"));
+        assertEquals("NumericData-01", e.getCondition().getDataId());
```
pilhuhn, line 190: _I_ would prefer to have such refactorings in a separate PR

bennetelli: Me too usually. Last time I opened a new PR because of refactorings like that, @josejulio said I should add it in an existing one :-D

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/AccountLatestUpdate.java
```diff
@@ -10,9 +10,13 @@
 @Table(name = "account_latest_update")
 public class AccountLatestUpdate {
 
+    // TODO POL-650 shouldn't we use auto generated ids as PK instead of natural keys?
     @Id
```
pilhuhn, line 14: There should only be one entry per account, so account-id as PK sounds correct to me.

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineImpl.java
```diff
@@ -45,6 +48,11 @@
 @ApplicationScoped
 public class LightweightEngineImpl implements LightweightEngine {
 
+    public static final String USE_ORG_ID = "policies.use-org-id";
```
pilhuhn, line 51: Use the one from NotificationActionPluginListener instead? This way a potential rename would only need to be done once

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
```diff
@@ -20,6 +21,11 @@
 @ApplicationScoped
 public class TriggersLoader {
 
+    public static final String USE_ORG_ID = "policies.use-org-id";
```
pilhuhn, line 24: Same as above.

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
```diff
@@ -47,30 +53,58 @@ public List<FullTrigger> getTriggers(String accountId) {
...
-        try {
-            return statelessSessionFactory.getCurrentSession().createQuery(hql, LocalDateTime.class)
-                    .setParameter("accountId", accountId)
-                    .getSingleResult();
-        } catch (NoResultException e) {
-            LOGGER.debugf("No latest triggers update found for account %s", accountId);
-            return LocalDateTime.MIN;
+
+        if (useOrgId) {
+            String hql = "SELECT latest FROM AccountLatestUpdate WHERE orgId = :accountId";
```
pilhuhn, line 58: Rename hql parameter to `:orgId`

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
```diff
@@ -47,30 +53,58 @@ public List<FullTrigger> getTriggers(String accountId) {
...
+                        .getSingleResult();
+            } catch (NoResultException e) {
+                LOGGER.debugf("No latest triggers update found for account %s", accountId);
+                return LocalDateTime.MIN;
+            }
+        } else {
+            String hql = "SELECT latest FROM AccountLatestUpdate WHERE accountId = :accountId";
+            try {
+                return statelessSessionFactory.getCurrentSession().createQuery(hql, LocalDateTime.class)
+                        .setParameter("orgId", accountId)
```
pilhuhn, line 71: Here the 1st param of `setParameter` should be `accountId` as defined in Line 68

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
```diff
@@ -47,30 +53,58 @@ public List<FullTrigger> getTriggers(String accountId) {
...
-        List<FullTrigger> triggers = statelessSessionFactory.getCurrentSession().createQuery(hql, Policy.class)
-                .setParameter("accountId", accountId)
-                .getResultList()
-                .stream()
-                .map(PolicyToTriggerConverter::convert)
-                .collect(toList());
-        for (FullTrigger trigger : triggers) {
-            provideDefaultDampening(trigger);
+        if (useOrgId) {
+            String hql = "SELECT p FROM Policy p WHERE p.orgId = :accountId";
```
pilhuhn, line 82: Rename `:accountId` to `:orgId` and use it in line 84

pilhuhn (changes requested):
I think there is one bug in TriggerLoader around line 71. Other comments are more cosmetics from my PoV

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/AccountLatestUpdate.java
bennetelli: Ok. A matter of taste maybe. That makes some use cases a bit harder than they could be.

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
bennetelli: this is only a semantical thing, right? The left part is the property of the entity and the right on the parameter.

bennetelli:
> I think there is one bug in TriggerLoader around line 71. Other comments are more cosmetics from my PoV

good catch. Thanks. Changed now.

gwenneg (changes requested):
I'll help fixing this PR after #380 is merged (which is about to happen). I'll review it as well after that.

.rhcicd/run_org_id_populator.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdJobInvocation
+metadata:
+  name: populate-org-id-column
+spec:
+  appName: notifications-backend
+  jobs:
+    - org-id-populator
+    - org-id-populator-event-table
```
bennetelli: that job is obsolete. Deleted.

gwenneg:
> Uh no :-D I spent one day of merging already.

Don't worry, I'll take care of it 😃

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/AccountLatestUpdate.java
bennetelli: Removed

engine/src/test/java/org/hawkular/alerts/engine/RulesEngineTest.java
bennetelli: > > Uh no :-D I spent one day of merging already.
> 
> Don't worry, I'll take care of it smiley

Ok :) If you need help, don't hesitate to raise your hand. I could help as well

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineImpl.java
gwenneg: [Bullet-proof approach](https://github.com/RedHatInsights/policies-engine/blob/master/external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineConfig.java): inject the config value in an `@ApplicationScoped` bean and inject that bean everywhere. It even makes the `application.properties` entry optional.

bennetelli: The entry should actually be optional already. At least I wanted it to be like that by adding the defaultValue.

bennetelli: @gwenneg added. But I left the entry in the application.properties. Maybe I am a bit too shy here

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineImpl.java
```diff
@@ -156,7 +160,10 @@ public void process(Event event) {
             // In the old implementation, the time comes from the Action constructor which relies on System.currentTimeMillis().
             LocalDateTime nowUTC = LocalDateTime.ofInstant(Instant.ofEpochMilli(System.currentTimeMillis()), UTC);
 
-            PoliciesAction policiesAction = new PoliciesAction();
+            PoliciesAction policiesAction = new PoliciesAction(orgIdConfig.isUseOrgId());
+
+            policiesAction.setOrgId(event.getTenantId());
```
bennetelli, line 165: should this be set based on the orgIdConfig entry as well?

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/LightweightEngineImpl.java
bennetelli: I am not into the depth of the framework internals, but I am expecting that they take the value during application start. Either from application.properties or environment variable, then "save" it and inject it everywhere where it's used without checking the status during that process. But that could be wrong.

external/src/main/java/com/redhat/cloud/policies/engine/lightweight/TriggersLoader.java
pilhuhn: Here is is cosmetics.

bennetelli: That ui is still broken. I changed that code already, but it doesn't show there were changes. And the "viewed" checkbox bug is also still there, Gwenneg mentioned this mornin

gwenneg:
⚠️🚧⚠️ Rebasing in progress, please do not push any new commit ⚠️🚧⚠️

bennetelli:
> Rebase done. However, after a quick review I think there are a few things that are missing or need to be changed.
> 
> Here's how I think the orgId migration should impact policies-engine: image
> 
> This diagram assumes that the `orgId` will come from the inventory Kafka message. Is that a correct assumption? If so, the `orgId` should be passed between policies-engine classes using a new `Event` field.
> 
> From a DB perspective, I think we should not change the existing `account_latest_update` table but rather create another similar `org_latest_update` table. The former will be dropped when the migration is over.

I changed it to a separate table.

src/main/java/com/redhat/cloud/policies/engine/db/entities/AccountLatestUpdate.java
```diff
@@ -14,5 +14,8 @@ public class AccountLatestUpdate {
     @Column(name = "account_id")
     public String accountId;
 
+    @Column(name = "org_id")
+    public String orgId;
```
gwenneg, line 18: This field is no longer needed here since tables are now separated. We need a new `OrgIdLatestUpdate` entity now.

src/test/resources/db/migration/V4__POL-550_ebs_accountnumber_to_org_id.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE org_id_latest_update (
```
gwenneg, line 1: Shouldn't the file name start with `V4__POL-650`? 😃

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
```diff
@@ -1,27 +1,13 @@
...
-import org.antlr.v4.runtime.atn.PredictionContextCache;
+import org.antlr.v4.runtime.atn.*;
 import org.antlr.v4.runtime.dfa.DFA;
-import org.antlr.v4.runtime.tree.ParseTreeListener;
-import org.antlr.v4.runtime.tree.ParseTreeVisitor;
-import org.antlr.v4.runtime.tree.TerminalNode;
-
+import org.antlr.v4.runtime.*;
+import org.antlr.v4.runtime.misc.*;
+import org.antlr.v4.runtime.tree.*;
```
gwenneg, line 7: Star imports are back ⁉️

src/main/java/com/redhat/cloud/policies/engine/process/EventProcessor.java
```diff
@@ -59,7 +63,10 @@ public void process(Event event) {
         } else {
             LOGGER.debugf("Found %d enabled policies for account %s", enabledPolicies.size(), event.getAccountId());
 
-            PoliciesAction policiesAction = new PoliciesAction();
+            PoliciesAction policiesAction = new PoliciesAction(orgIdConfig.isUseOrgId());
+
+            policiesAction.setOrgId(event.getAccountId());
```
gwenneg, line 68: This should be:
```suggestion
            policiesAction.setOrgId(event.getOrgId());
```
but first Event needs to be updated with the `orgId` value retrieved in `PayloadParser`.

src/main/java/com/redhat/cloud/policies/engine/process/PoliciesAction.java
```diff
@@ -21,12 +21,27 @@
...
+    }
+
     public String getKey() {
-        return accountId + context.systemCheckIn;
+        if (useOrgId) {
+            return orgId + context.systemCheckIn;
+        } else {
+            return accountId + context.systemCheckIn;
+        }
     }
```
gwenneg, line 45: This method isn't used anymore, it's a relic from the past. It can be deleted.

As a consequence, I don't think we need to pass the `useOrgId` boolean to the constructor.

src/main/java/com/redhat/cloud/policies/engine/process/EventProcessor.java
```diff
@@ -59,7 +64,12 @@ public void process(Event event) {
         } else {
             LOGGER.debugf("Found %d enabled policies for account %s", enabledPolicies.size(), event.getAccountId());
 
-            PoliciesAction policiesAction = new PoliciesAction();
+            PoliciesAction policiesAction = new PoliciesAction(orgIdConfig.isUseOrgId());
+
+            if (orgIdConfig.isUseOrgId()) {
+                policiesAction.setOrgId(event.getOrgId());
```
gwenneg, line 70: Same thing here, setting the `orgId` all the time will never cause an issue even if it's missing. It doesn't need to be enabled from configuration.

src/main/java/com/redhat/cloud/policies/engine/process/PoliciesAction.java
bennetelli: But the getKey() method would not work anymore when there is no flag present

bennetelli: this one: https://github.com/RedHatInsights/policies-engine/pull/377/files#diff-54a89ce45518be890cf1a0c32d4002540420853123cbf94b647e5fb638bcd4dcR39

gwenneg: Yes but it's not used anymore. The method can be safely deleted.

src/test/resources/db/migration/V4__POL-550_ebs_accountnumber_to_org_id.sql
bennetelli: indeed

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
bennetelli: wtf.. we do have editorconfig?!

bennetelli: ah. the code generation sh** is doing this automagically

bennetelli:
@gwenneg it's getting a bit confusing with the threads. I would say I fixed every requested change

src/main/java/com/redhat/cloud/policies/engine/process/Event.java
```diff
@@ -27,6 +28,16 @@ public Event(String accountId, String id, String category, String text) {
...
+    public static Event createEventWithOrgId(String orgId, String id, String categoryName, String text) {
+        Event event = new Event();
+        event.setOrgId(orgId);
+        event.setId(id);
+        event.setCategory(categoryName);
+        event.setText(text);
+
+        return event;
+    }
+
```
gwenneg, line 40: This is not needed anymore:
```suggestion
```

src/main/java/com/redhat/cloud/policies/engine/process/Event.java
```diff
@@ -108,6 +119,14 @@ public void setContext(Map<String, String> context) {
...
+        return orgId;
+    }
+
+    public void setOrgId(String orgId) {
+        this.orgId = orgId;
+    }
+
     @Override
     public String toString() {
         return "Event [accountId=" + accountId + ", id=" + id + ", ctime=" + ctime + ", category=" + category
```
gwenneg, line 132:
```suggestion
        return "Event [accountId=" + accountId + ", orgId=" + orgId + ", id=" + id + ", ctime=" + ctime + ", category=" + category
```

src/main/resources/application.properties
```diff
@@ -81,3 +81,5 @@ quarkus.hibernate-orm.physical-naming-strategy=com.redhat.cloud.policies.engine.
 quarkus.http.non-application-root-path=/
 
 quarkus.cache.caffeine.account-latest-update.expire-after-write=PT1M
+
+policies.use-org-id=false
```
gwenneg, line 85: I doubt this will ever be modified. We'll change the value from an env var.
```suggestion
```

src/main/java/com/redhat/cloud/policies/engine/db/entities/Policy.java
```diff
@@ -16,6 +16,8 @@ public class Policy {
     @Column(name = "customerid")
     public String accountId;
 
+    public String orgId;
```
gwenneg, line 19: This field should be added to the SQL schema in [V4__POL-650_ebs_accountnumber_to_org_id.sql](https://github.com/RedHatInsights/policies-engine/pull/377/files#diff-b92d6ca9912131efbf9117f86818d4de71d4f28e6637283659b781694bcdadd1)

.rhcicd/run_org_id_populator.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdJobInvocation
+metadata:
+  name: populate-org-id-column
+spec:
+  appName: notifications-backend
```
gwenneg, line 7: Shouldn't this be `policies` instead of `notifications`?

.rhcicd/clowdapp.yaml
```diff
@@ -18,6 +18,37 @@ objects:
     - policies-ui-backend
     database:
       sharedDbAppName: policies-ui-backend
+    jobs:
+    - name: org-id-populator
```
gwenneg, line 22: The `policies-ui-backend` app is supposed to own the `policies-ui-backend` DB and it should also take care of migrating its data.

Shouldn't this job by in `policies-ui-backend`?

src/main/java/com/redhat/cloud/policies/engine/config/OrgIdConfig.java
```diff
@@ -0,0 +1,23 @@
+package com.redhat.cloud.policies.engine.config;
```
gwenneg, line 1: Wrong file but I can't comment the good one: this PR adds an empty [ExpressionLexer.java](https://github.com/RedHatInsights/policies-engine/pull/377/files#diff-cbdcb24a2f2b3470d91ced2bb5f251c5516c3074f1f9bed7cd07318022ed98bd) file which should be removed.

src/main/java/com/redhat/cloud/policies/engine/process/PoliciesAction.java
```diff
@@ -21,14 +21,11 @@
 
 public class PoliciesAction {
     private String accountId;
+    private String orgId;
```
gwenneg, line 24: This needs to be tested in `NotificationSenderTest`.

src/main/java/com/redhat/cloud/policies/engine/db/repositories/PoliciesRepository.java
```diff
@@ -44,22 +48,36 @@ public List<Policy> getEnabledPolicies(String accountId) {
...
         LOGGER.debugf("Finding latest policies update time for account %s", accountId);
-        String hql = "SELECT latest FROM AccountLatestUpdate WHERE accountId = :accountId";
         try {
-            return statelessSessionFactory.getCurrentSession().createQuery(hql, LocalDateTime.class)
-                    .setParameter("accountId", accountId)
-                    .getSingleResult();
+            if (orgIdConfig.isUseOrgId()) {
+                String hql = "SELECT latest FROM OrgIdLatestUpdate WHERE orgId = :orgId";
+                return statelessSessionFactory.getCurrentSession().createQuery(hql, LocalDateTime.class)
+                        .setParameter("orgId", accountId)
```
gwenneg, line 55: The `orgId` should be used here, not the `accountId`. As I mentioned earlier, it would be easier to duplicate the `getEnabledPolicies` method and pass the `orgId` to the new method. The `orgIdConfig.isUseOrgId()` condition would determine which method should be called.

src/main/java/com/redhat/cloud/policies/engine/db/repositories/PoliciesRepository.java
```diff
@@ -44,22 +48,36 @@ public List<Policy> getEnabledPolicies(String accountId) {
...
 
     private List<Policy> findEnabledPoliciesByAccount(String accountId) {
-        String hql = "FROM Policy WHERE accountId = :accountId AND enabled IS TRUE";
-        return statelessSessionFactory.getCurrentSession().createQuery(hql, Policy.class)
-            .setParameter("accountId", accountId)
-            .getResultList();
+        if (orgIdConfig.isUseOrgId()) {
+            String hql = "FROM Policy WHERE orgId = :orgId AND enabled IS TRUE";
+            return statelessSessionFactory.getCurrentSession().createQuery(hql, Policy.class)
+                    .setParameter("orgId", accountId)
```
gwenneg, line 73: Same here, the value should be `orgId`. Same solution suggestion: duplicate the method and check `orgIdConfig.isUseOrgId()` to determine which method should be called.

gwenneg:
`PoliciesHistoryEntry` and `PoliciesHistoryRepository` will need the `orgId` as well.

.rhcicd/run_org_id_populator.yaml
bennetelli: afaik it doesn't really matter and will be deleted anyway, but yes. Copy paste issue

bennetelli:
Added every change you requested, @gwenneg

.rhcicd/clowdapp.yaml
```diff
@@ -123,3 +123,11 @@ parameters:
...
   description: Enable Sentry (or not)
   value: "false"
+- description: Host for the EAN to OrgId translator.
+  name: TENANT_TRANSLATOR_HOST
+  required: true
+  value: 'apicast.3scale-dev.svc.cluster.local'
+- description: Port for the EAN to OrgId translator.
+  name: TENANT_TRANSLATOR_PORT
+  required: true
+  value: '8892'
```
gwenneg, line 133: Is this still needed now that the populator job is no longer in this PR?

bennetelli: no it's not. removed

.rhcicd/run_org_id_populator.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: cloud.redhat.com/v1alpha1
+kind: ClowdJobInvocation
+metadata:
+  name: populate-org-id-column
+spec:
+  appName: policies-engine
+  jobs:
+    - org-id-populator
```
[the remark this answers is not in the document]
bennetelli: no it's not. removed

bennetelli:
That should be it. Hopefully. @gwenneg

src/test/java/com/redhat/cloud/policies/engine/process/EventProcessorTest.java
```diff
@@ -125,6 +125,7 @@ void testEnabledPoliciesFoundAndOnePolicyFiredButNotSent() {
     private static Event buildEvent() {
         Event event = new Event();
         event.setAccountId("account-id");
+        event.setOrgId("org-id");
```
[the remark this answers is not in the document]
bennetelli: Added Nested classes in the test class so it's easier to detect and better separated from the accountId tests

src/main/java/com/redhat/cloud/policies/engine/db/repositories/PoliciesRepository.java
```diff
@@ -41,30 +43,68 @@ public List<Policy> getEnabledPolicies(String accountId) {
...
+    }
+
+    private List<Policy> findEnabledPoliciesByOrgId(String orgId) {
+        String hql = "FROM Policy WHERE orgId = :orgId AND enabled IS TRUE";
+        return statelessSessionFactory.getCurrentSession().createQuery(hql, Policy.class)
+                .setParameter("orgId", orgId)
+                .getResultList();
     }
 
     @CacheInvalidateAll(cacheName = ACCOUNT_LATEST_UPDATE_CACHE_NAME)
```
bennetelli, line 99: @gwenneg do we (you) still need that cache?

bennetelli: argh again. I meant the clear cache method which seems to be unused.

src/main/java/com/redhat/cloud/policies/engine/db/repositories/PoliciesRepository.java
```diff
@@ -41,30 +43,68 @@ public List<Policy> getEnabledPolicies(String accountId) {
...
     public void clearAllCaches() {
```
bennetelli, line 100: that's the line I wanted to mark and ask if we need these two methods, @gwenneg

src/main/java/com/redhat/cloud/policies/engine/db/repositories/PoliciesRepository.java
gwenneg: Hmm, it was used from a test initially. Maybe that test was changed and this method is no longer needed.

bennetelli:
> LGTM

YAY :)

PR #380, opened 2022-05-16
POL-649 Remove Hawkular

gwenneg:
⚠️ WIP

I'll try to break that down into several "smaller" PRs.


external/src/main/java/com/redhat/cloud/policies/engine/process/LightweightEngine.java
```diff
@@ -0,0 +1,206 @@
...
+import com.redhat.cloud.notifications.ingress.Payload;
+import com.redhat.cloud.policies.api.model.condition.expression.ExprParser;
+import com.redhat.cloud.policies.engine.db.repositories.PoliciesRepository;
+import com.redhat.cloud.policies.engine.db.entities.Policy;
+import com.redhat.cloud.policies.engine.db.repositories.PoliciesHistoryRepository;
+import io.smallrye.reactive.messaging.kafka.api.OutgoingKafkaRecordMetadata;
+import io.vertx.core.json.JsonObject;
+import org.apache.kafka.common.header.internals.RecordHeaders;
+import org.eclipse.microprofile.metrics.Counter;
+import org.eclipse.microprofile.metrics.annotation.Metric;
```
bennetelli, line 16: we should migrate to micrometer here as well in a later PR

external/src/main/java/com/redhat/cloud/policies/engine/process/LightweightEngine.java
```diff
@@ -0,0 +1,206 @@
...
+                    emitter.send(buildMessageWithId(payload));
+                    notificationsCounter.inc();
+                } catch (IOException e) {
+                    LOGGER.warnf("Failed to serialize action for accountId", policiesAction.getAccountId());
+                }
+            }
+        }
+    }
+
+    private boolean shouldSendNotification(Policy policy) {
```
bennetelli, line 161: maybe just "sendNotification"? I am using should very often in tests so it's clear what to expect from a test. 
but I am also fine with the current name. Do it as you like :)

external/src/main/resources/application.properties
```diff
@@ -22,12 +18,8 @@ quarkus.http.access-log.enabled=true
 quarkus.http.access-log.category=access_log
 quarkus.http.access-log.pattern=combined
 
-
-# For dev only
-#external.org.hawkular.alerts.ui.path=/home/michael/projects/hawkular-alerts-ui
-
 # Kafka bootstrap applies to all topics
-kafka.bootstrap.servers=localhost:9092
+kafka.bootstrap.servers=localhost:29092
```
bennetelli, line 22: why do we need to change the port as well?

.github/workflows/maven.yml
```diff
@@ -20,4 +20,4 @@ jobs:
         java-version: 11
         distribution: adopt
     - name: Build with Maven
-      run: ./mvnw -B clean install -Pitest -Dhawkular.data=target/hawkular.data --no-transfer-progress
+      run: ./mvnw -B clean install -Pitest --no-transfer-progress
```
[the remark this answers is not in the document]
gwenneg: `verify` is enough indeed, changed.

external/src/main/resources/application.properties
gwenneg: Fixed.

external/src/main/java/com/redhat/cloud/policies/engine/process/LightweightEngine.java
gwenneg: I don't know, `sendNotification` sounds like the method will actually do that while it won't.

bennetelli: true. Naming is hard.. I also don't know a better name. so leave it as it is.

gwenneg:
image

external/src/main/java/com/redhat/cloud/policies/engine/RunOnStartup.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.regex.Pattern;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.annotation.PostConstruct;
+import org.jboss.logging.Logger;
+
+@Startup
+public class RunOnStartup {
+
+    // TODO POL-649 Update that pattern after the new health check call has been confirmed.
```
josejulio, line 16: I'm supposing this is to ease the updating to the new API ?

external/src/main/java/com/redhat/cloud/policies/engine/process/EventProcessor.java
```diff
@@ -0,0 +1,127 @@
...
+import static com.redhat.cloud.policies.engine.process.PayloadParser.DISPLAY_NAME_FIELD;
+import static com.redhat.cloud.policies.engine.process.PayloadParser.INVENTORY_ID_FIELD;
+import static java.time.ZoneOffset.UTC;
+
+@ApplicationScoped
+public class EventProcessor {
+
+    private static final Logger LOGGER = Logger.getLogger(EventProcessor.class);
+    private static final Pattern ACTIONS_SPLIT_PATTERN = Pattern.compile(";");
+    private static final List<String> NOTIFICATIONS_ACTIONS = List.of("email", "notification");
```
josejulio, line 31: I guess this could be simplified to only notifications (gabi might confirm). At some point we did migrate the SQL from `email` to `notification`- This was added because we didn't dare to migrate the infinispan data :grin:

gwenneg: It could indeed. I'll check with gabi first when I'll have access to the policies DB.

external/src/main/java/com/redhat/cloud/policies/engine/RunOnStartup.java
gwenneg: Yes, this can be done later.

Before the lightweight engine, `ui-backend` called the `engine` getTrigger REST endpoint as part of its health check. I replaced that with the only remaining `engine` REST endpoint with is `/validate`. But I don't really like that. I think we should have a dedicated `engine` endpoint for the `ui-backend` health check. It could be something as simple as `GET /ping`. The pattern here will need an update after we're done chosing the best REST call.

gwenneg: Filtering all `/validate` calls wouldn't make sense as some of them will be real requests that don't come from the health check.

external/src/test/java/com/redhat/cloud/policies/engine/process/NotificationSenderTest.java
```diff
@@ -0,0 +1,148 @@
...
+        assertEquals(EVENT_TYPE_NAME, action.getEventType());
+        assertEquals(policiesAction.getAccountId(), action.getAccountId());
+        assertEquals(policiesAction.getTimestamp(), action.getTimestamp());
+        assertEquals(policiesAction.getContext().getInventoryId(), action.getContext().getAdditionalProperties().get("inventory_id"));
+        assertEquals(policiesAction.getContext().getDisplayName(), action.getContext().getAdditionalProperties().get("display_name"));
+
+        // The system_check_in field precision is different before and after the action serialization.
+        LocalDateTime expectedSystemCheckIn = policiesAction.getContext().getSystemCheckIn();
+        String actualSystemCheckIn = (String) action.getContext().getAdditionalProperties().get("system_check_in");
+        assertTrue(expectedSystemCheckIn.toString().startsWith(actualSystemCheckIn));
```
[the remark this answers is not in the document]
gwenneg: I pushed something close. Thanks, that looks better 😃

PR #385, opened 2022-05-19
POL-649 Remove PV claim and lower resources limits

gwenneg:
This PR:
- removes all ClowdApp configuration related to the persistent volume
- slightly reduces the probes initial delay
- lowers the CPU/memory requests and limits

⚠️ It may cause the removal of the PV and all its data at deployment time. I am not sure whether that will happen automatically or if a human action (from devprod) will be needed to delete it.

### DIFF
--- a/.rhcicd/clowdapp.yaml
+++ b/.rhcicd/clowdapp.yaml
@@ -4,18 +4,6 @@ kind: Template
 metadata:
   name: policies-engine
 objects:
-- apiVersion: v1
-  kind: PersistentVolumeClaim
-  metadata:
-    name: policies-engine-data
-    labels:
-      app: policies-engine
-  spec:
-    accessModes:
-    - ReadWriteOnce
-    resources:
-      requests:
-        storage: ${INFINISPAN_STORAGE}
 - apiVersion: cloud.redhat.com/v1alpha1
   kind: ClowdApp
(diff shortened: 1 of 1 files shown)


.rhcicd/clowdapp.yaml
```diff
@@ -129,19 +110,15 @@ parameters:
...
   description: Memory limit
-  value: 3Gi
+  value: 1Gi
 - name: MEMORY_REQUEST
   description: Memory request
-  value: 1.5Gi
-- name: REPLICAS
+  value: 500Mi
+- name: MIN_REPLICAS
   value: "1"
```
bennetelli, line 122: shouldn't this be something uneven? 3 at least?

bennetelli (changes requested):
shouldn't we have something uneven for the min replicas? 3 at least?

.rhcicd/clowdapp.yaml
gwenneg: We usually have 1 replica in the ClowdApp template and that value is increased from app-interface depending on the target environment.

bennetelli: ah. you are right. ok. then you get my lgtm

bennetelli (approved):

PR #386, opened 2022-05-19
POL-649 Merge Maven modules

gwenneg:
This is the last major PR related to [POL-649](https://issues.redhat.com/browse/POL-649).

It moves all the code from the old `api` and `external` modules to the root of the project and deletes these modules.

It also contains an update of the conditions parsing code generated by antlr `4.10.1`. Apparently we're supposed to regenerate that code when the antlr runtime version is bumped. Otherwise, that dependency will complain in the logs. I think we didn't see it before because the latest version was automatically generated each time the Docker image was built. It doesn't hurt to commit the updated files anyway.


pom.xml
```diff
@@ -25,115 +23,245 @@
...
-    <quarkus.version>2.9.1.Final</quarkus.version>
+    <!-- Plugins -->
+    <compiler-plugin.version>3.10.1</compiler-plugin.version>
+    <surefire-plugin.version>3.0.0-M6</surefire-plugin.version>
 
-    <!--Distributed tracing support for Vert.x which is not covered by SmallRye-OT -->
-    <version.opentracing-vertx-web>1.0.0</version.opentracing-vertx-web>
+    <!-- Quarkus -->
+    <quarkus.platform.artifact-id>quarkus-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus.platform</quarkus.platform.group-id>
```
bennetelli, line 40: maybe a stupid question, but does the groupId and artifactId change that often in Quarkus so we need it to be replaceable?

pom.xml
```diff
@@ -25,115 +23,245 @@
...
+    </dependency>
+
+    <!-- Misc -->
+    <dependency>
+      <groupId>org.eclipse.microprofile.metrics</groupId>
+      <artifactId>microprofile-metrics-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
```
bennetelli, line 152: should we add a task to replace the MultiMap we are using with plain java and get rid of this dependency?

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
```diff
@@ -963,32 +963,66 @@ private boolean object_sempred(ObjectContext _localctx, int predIndex) {
...
-		"\29:\5\f\7\2:;\5\24\13\2;I\3\2\2\2<=\5\32\16\2=>\5\16\b\2>?\5\22\n\2?"+
-		"I\3\2\2\2@A\5\32\16\2AB\5\16\b\2BC\5\26\f\2CI\3\2\2\2DE\5\32\16\2EF\5"+
-		"\20\t\2FG\5\22\n\2GI\3\2\2\2H\63\3\2\2\2H\64\3\2\2\2H8\3\2\2\2H<\3\2\2"+
-		"\2H@\3\2\2\2HD\3\2\2\2I\7\3\2\2\2JK\t\2\2\2K\t\3\2\2\2LM\t\3\2\2M\13\3"+
-		"\2\2\2NO\t\4\2\2O\r\3\2\2\2PQ\t\5\2\2Q\17\3\2\2\2RS\t\6\2\2S\21\3\2\2"+
-		"\2TU\7\5\2\2UZ\5\26\f\2VW\7\6\2\2WY\5\26\f\2XV\3\2\2\2Y\\\3\2\2\2ZX\3"+
-		"\2\2\2Z[\3\2\2\2[]\3\2\2\2\\Z\3\2\2\2]^\7\7\2\2^b\3\2\2\2_`\7\5\2\2`b"+
-		"\7\7\2\2aT\3\2\2\2a_\3\2\2\2b\23\3\2\2\2cd\t\7\2\2d\25\3\2\2\2ef\t\b\2"+
-		"\2f\27\3\2\2\2gh\t\t\2\2h\31\3\2\2\2ij\t\n\2\2j\33\3\2\2\2\7(\60HZa";
+		"\u0004\u0001\u0019j\u0002\u0000\u0007\u0000\u0002\u0001\u0007\u0001\u0002"+
```
bennetelli, line 966: unfortunately I don't understand that change.

pom.xml
gwenneg: Yeah, we should definitely get rid of Guava.

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
gwenneg: It's auto-generated code. I don't know what that is but antlr made that change 😃

pom.xml
gwenneg: Even if we needed to change that (which doesn't happen often), we could still do it in the dependency declaration. All Quarkus projects are generated that way but these properties are a bit useless indeed.

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
bennetelli: then it's fine :-)

pom.xml
bennetelli: Added a jira for it: https://issues.redhat.com/browse/POL-670

src/main/java/com/redhat/cloud/policies/engine/condition/ExpressionParser.java
bennetelli: then it's probably fine as it is :-D

pom.xml
bennetelli: then let's just keep it. I was just interested in the reason why we/quarkus is doing this

bennetelli (approved):

PR #388, opened 2022-05-19
Use non-deprecated naming strategy

bennetelli:


gwenneg (approved):
Rebased on master.

LGTM

PR #392, opened 2022-05-23
POL-650 Add orgId to PoliciesAction

gwenneg:
Same as [policies-engine#377](https://github.com/RedHatInsights/policies-engine/pull/377) with conflicts solved.

⚠️ Self-review in progress, I need to make sure I didn't screw up during the merge 😄


src/main/java/com/redhat/cloud/policies/engine/process/PoliciesAction.java
```diff
@@ -21,12 +21,27 @@
...
+    private String orgId;
     private LocalDateTime timestamp;
     private Context context = new Context();
     private Set<Event> events = new HashSet<>();
 
+    private final boolean useOrgId;
+
+    public PoliciesAction() {
+        this(false);
+    }
```
gwenneg, line 33: This wasn't in the original PR.

gwenneg:
@bennetelli I double-checked every change in this PR, it is equivalent to #377 (with one exception commented above). Is that OK with you if I push this branch onto the existing changes in #377?

bennetelli:
> @bennetelli I double-checked every change in this PR, it is equivalent to #377 (with one exception commented above). Is that OK with you if I push this branch onto the existing changes in #377?

sure it is. You could even just do it and let me know afterwards :)

gwenneg:
Pushed onto the other branch.

PR #394, opened 2022-05-25
Decrease default CPU request/limit

gwenneg:
Metrics show that the current CPU request and limit are way above the actual CPU used by the containers on prod.


gwenneg:
image

theute (approved):

PR #405, opened 2022-06-15
Use the new Quarkus Log API

gwenneg:
Quarkus recently introduced a new `Log` API which delegates all its invocations to `org.jboss.logging.Logger`. Building the logger is no longer needed.


theute (approved):
Pretty !

PR #422, opened 2022-07-16
Make build more reproducible.

pilhuhn:


pom.xml
```diff
@@ -25,15 +25,22 @@
...
-    <maven.compiler.target>11</maven.compiler.target>
+    <maven.compiler.source>$java-version}</maven.compiler.source>
+    <maven.compiler.target>$java-version}</maven.compiler.target>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <!-- Plugins -->
     <compiler-plugin.version>3.10.1</compiler-plugin.version>
     <surefire-plugin.version>3.0.0-M7</surefire-plugin.version>
+    <maven-jar-plugin.version>3.2.2</maven-jar-plugin.version>
```
gwenneg, line 43: Is this needed? We stopped producing jars with the new policies.

pilhuhn: The jar plugin and others are always there, but defined in the Super-POM within the used maven distro.
Explicitly setting this, prevents the dependency checker from complaining.
I agree that we don't strictly need it, but I think it also does no harm.

gwenneg (approved):
LGTM

PR #517, opened 2023-04-21
Fix missing details for cloud events

josejulio:
- Setting the base env url in clowdapp
 - Updating event-schemas version
    - Relax validation for description
    - Correctly decode /encode the date-time
    - policy ids are uuids


gwenneg (approved):
LGTM

PR #523, opened 2023-04-28
RHCLOUD-24299 Protect against Kafka poison pills

gwenneg:
The Policies engine isn't protected enough against [Kafka poison pills](https://medium.com/lydtech-consulting/kafka-poison-pill-e146b87c1866). This PR adds the most basic yet very efficient way of preventing that kind of issue. You may want to use a different [error handling strategy](https://quarkus.io/guides/kafka#error-handling) in the future but in the meantime I recommend that this PR is merged as a first step.

### DIFF
--- a/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
+++ b/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
@@ -32,10 +32,14 @@ public class Receiver {
     @ActivateRequestContext
     public void process(String payload) {
         Log.tracef("Received payload: %s", payload);
-        payloadParser.parse(payload).ifPresent(event -> {
-            statelessSessionFactory.withSession(statelessSession -> {
-                eventProcessor.process(event);
+        try {
+            payloadParser.parse(payload).ifPresent(event -> {
+                statelessSessionFactory.withSession(statelessSession -> {
+                    eventProcessor.process(event);
+                });
             });
-        });
+        } catch (Exception e) {
+            Log.errorf(e, "Payload processing failed: %s", payload);
+        }
     }
 }
--- a/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
+++ b/src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
@@ -39,7 +39,15 @@ public class Receiver {
                 });
             });
         } catch (Exception e) {
-            Log.errorf(e, "Payload processing failed: %s", payload);
+            if (Log.isDebugEnabled()) {
+                /*
+                 * When the DEBUG log level is enabled, the log entry will include the payload.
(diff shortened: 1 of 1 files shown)


src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
```diff
@@ -32,10 +32,14 @@ public class Receiver {
...
-                eventProcessor.process(event);
+        try {
+            payloadParser.parse(payload).ifPresent(event -> {
+                statelessSessionFactory.withSession(statelessSession -> {
+                    eventProcessor.process(event);
+                });
             });
-        });
+        } catch (Exception e) {
+            Log.errorf(e, "Payload processing failed: %s", payload);
```
vkrizan, line 42: Payload might be too big for a log message.

vkrizan (approved):
Thank you

src/main/java/com/redhat/cloud/policies/engine/process/Receiver.java
gwenneg: Good point. I don't think that will ever be an issue for the OpenShift log, but it may cause a failure when a batch of log entries is pushed to CloudWatch.

Do you want to truncate the payload when it exceeds a chosen max length (which one)?

vkrizan: Can the error contain the error message? We should leave out payload to debug data/log.

gwenneg: I updated the code. The log entry depends on the log level now.

gwenneg: The log entry ar `error` level will contain all useful details about the exception.

gwenneg: The `e` argument which is passed to `Log.error[f]` method will trigger a whole stack trace print in the logs. That includes the exception message so there's no need to repeat it in the message passed to `Log.error[f]`.

gwenneg: With the current code, the exception message and the full stack trace are printed whether the debug log is enabled or not. `e` is passed in both cases but in one case it's the first arg and in the other case it's the last arg. I don't know why they built the API like that, it can be confusing...

vkrizan: With the `e` passed as second arg, it is taken as a param for the format (passed as first). See  [`errorf(java.lang.String,java.lang.Object)`](https://javadoc.io/static/io.quarkus/quarkus-core/2.7.2.Final/io/quarkus/logging/Log.html#errorf(java.lang.String,java.lang.Object)). Are you sure it would be printed when the format does not have a placeholder?

PR #559, opened 2023-10-03
Make Kafka config init compatible with MSK

gwenneg:
This should help fix the `policies-engine` deployment in stage.

I made the [same kind of change](https://github.com/RedHatInsights/notifications-backend/pull/2216) in Notifications and now everything works like a charm.


vkrizan (approved):
Thank you!

PR #591, opened 2024-01-23
feat(consumer): improved processing logging

vkrizan:
Improves processing logging by adding more debug logs, more information, and info logs for fired policies.

Besides that, it:
* re-throws JSON parse errors that later are captured and error logged 
* `Event` keeps recordKey (key from a Kafka message) -- for Inventory events this would be a host id
* `Event.id` is pre-generated as `UUID` and shared when producing a CloudEvent
* couple of code refactor to accommodate that

RHINENG-7089


petrblaho (approved):
LGTM!

PR #643, opened 2024-09-19
Convert container image/Dockerfile from UBI8 to UBI9

patchkez:
This PR should replace of using UBI8 images with UBI9.
I was able to build new image with:

```
podman build . -f ./src/main/docker/Dockerfile.jvm -t test_build_policies_ui_backend --ulimit=nofile=4096
```

I confirmed that curl exists in the final image as I had to remove it from dockerfile due to installation conflict.

### DIFF
--- a/src/main/docker/Dockerfile-build.jvm
+++ b/src/main/docker/Dockerfile-build.jvm
@@ -3,21 +3,21 @@
 ###
 
 # Build the project
-FROM registry.access.redhat.com/ubi8/openjdk-17:latest AS build
+FROM registry.access.redhat.com/ubi9/openjdk-17:latest AS build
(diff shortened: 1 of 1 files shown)


vkrizan:
@patchkez can you please also change the GitHub Action that does an image check? https://github.com/RedHatInsights/policies-engine/blob/master/.github/workflows/checkupdate.yml#L23

patchkez:
> @patchkez can you please also change the GitHub Action that does an image check? https://github.com/RedHatInsights/policies-engine/blob/master/.github/workflows/checkupdate.yml#L23

done

vkrizan (approved):
Thank you

PR #682, opened 2025-01-08
chore(cleanup): remove unused GC env var

vkrizan:
The env var GC_CONTAINER_OPTIONS is not being used on the ubi-minimal container.  I premume it was used with run-java.sh script that is no longer used for a while.


fellipeh (approved):
LGTM

PR #683, opened 2025-01-08
chore(cleanup): reduce heap on deployment

vkrizan:
After the Kafka OOM is fixed, getting back to the previou-ish values.

RHINENG-15124


fellipeh (approved):
lgtm

